PR #17, opened 2015-08-15
set ssb-config appname via $ssb_appname envvar.

dominictarr:
this makes it easy to test user onboarding.
```
cb patchwork
ssb_appname=foo electron .
```
and the database will be created under `~.{ssb_appname}`.


pfraze:
:+1: :+1: :+1:

PR #23, opened 2015-09-15
add dependencies details

mixmix:
I see @dominictarr has an open PullRequest which links out to details. 
I found it a little inconvenient and missed things first time (I still may be missing things because npm pack isn't working for me yet)

I also wonder if a recommended version of Node / IOJS would be nice here ... I saw references to a range of versions like `> Nodev0.12`, and  `= IOv2.5.0` . Would be good to know if there is anything tricky that needs to be known there

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,13 +8,47 @@ Patchwork isn't a website: it runs on your computer and syncs over the Internet 
 
 <a href="http://ssbc.github.io/">Learn more</a>
 
-## Some development notes:
+
+## Running Patchwork
+
+**Dependencies:**
+
+```
+# Linux dependencies
+apt-get install automake
+
+# OSX dependencies
+brew install automake
+```
+
+```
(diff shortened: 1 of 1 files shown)


pfraze:
Ah, yeah thanks. It's on my todo list (very soon) to get all these docs cleaned up.

Are nsis and wine dependencies? It's been long enough that I dont remember having to install those.

mixmix:
exactly!  super valuable having noobs like me to run through.

wine is a dependency if you want to make windows installers
nsis is some cullsoft installer script ... it implies that's needed in all
instances, but it could be just windows, not sure

https://github.com/loopline-systems/electron-builder/tree/v2.0.0

pfraze:
Ohhh gosh that's right, thanks for the reminder.

Yeah cool, really appreciate this! :star2:

PR #24, opened 2015-09-15
Backup

mixmix:
**[UPDATE]** - this pull request is just being kept as a reminder to include some easy backup guide/ feature for users

---

hey @pfraze , this is a question and a stab at an answer. 

Is what I've appended to the README correct? If I copy my `secret` to a new machine, can I resume my previous identity? Can I use patchwork on multiple computers with the same identity if I copy that secret across? 

I guess there's a later question about where the packaged patchwork store identity, but that's a question for later.


pfraze:
@mixmix yeah, you're correct that `~/.ssb/secret` is your private key, and that you want to keep a backup.

However, you definitely shouldn't use it only multiple machines. The problem is, you *never* want to publish two messages with the same sequence number. Why: peers won't know how to decide between them. They'll accept the first they're given, and reject any attempts to overwrite that. Then, you won't be able to add messages *after* that, because the `previous` hash (the blockchain) will not match.

mixmix:
that makes sense :)  I'm still thinking in centralised data servers
obviously

I guess the solution is a way to merge two identities. But not a pressing
need I imagine.

mixmix:
p.s.
are you seeing my messages in patchwork ok ? (could be a classic question
of not enough people on the network) or my patchy satellite link

p.p.s.
if my pull request isn't a good fit, please close it, I totally appreciate
the overhead of open-source (and really stoked about how fast you replied)

pfraze:
> that makes sense :)  I'm still thinking in centralised data servers obviously

Yeah that's one of those things, once you know it makes sense, but it's a surprising detail.

> I guess the solution is a way to merge two identities.

Yeah, declaring aliases and such

> are you seeing my messages in patchwork ok ?

yep yep, just replied

> if my pull request isn't a good fit, please close it,

Let's keep it open so I remember to put in all the info in this discussion

dominictarr:
@mixmix for clarity, it is okay to copy it to a new device, but do not use it from both devices. If you buy a new computer, it would be best to copy the entire `~/.ssb` directory, to ensure you get all the same posts that you had... otherwise if the sequence numbers are not the same you could create a new message at the wrong sequence.

This isn't implemented yet, but having two validly signed messages with the same sequence numbers will indicate that feed is compromised by the other peers...

PR #26, opened 2015-09-21
add .nvmrc for easy `nvm use`

mixmix:
nvm will check for this dotfile and use that for the version you're refering to. This lets you just call
`nvm install` or 
`nvm use`

without having to specify a version.
It has the added bonus of being a clear way to signal the preferred node version to use, which I like a lot


pfraze:
:+1: @mixmix :star:

mixmix:
please, I'm just polishing!

pfraze:
well I really appreciate it!

mixmix:
looking forward to helping out more, where possible

PR #29, opened 2015-09-29
this makes the wells have a nice paper-like feel

karissa:


karissa:
ignore, opening a real one

PR #43, opened 2015-10-17
osx: don’t quit on window close, just hide

davidguttman:


aredridel:
Oh, neat. I took a different approach with #45 -- I think a combination might even be the best.

pfraze:
This got done in the `patchwork-electron` module

PR #44, opened 2015-10-18
Update version for electron packager to 0.33.8

aredridel:


pfraze:
good catch, thanks /cc @dominictarr (you should see this)

PR #45, opened 2015-10-18
Add window hiding, remove window closing since there's no way to reopen it

aredridel:


pfraze:
Tested in OSX, looks good. Thanks @aredridel !

Issue #74, opened 2015-11-07
Improve feed load times


dominictarr:
I'm trying to figure out why loading that view would be so slow? it's strange because opening the main feed is fast, but opening the posts of a single author is slow. clue: it's fast for someone hasn't posted much, but noticably slower if they have posted lots...

For either me or @pfraze our posts load about 2 seconds on my computer (spinning harddrive)
but for substack it's noticably quicker, but still slower than someone who has only posted a few posts.

I did some investigation, of course I don't know that code very well, but I added a bunch of logging statements, and it looks like on the main feed it calls relatedMessages about 30 times (roughly a page and a half of logging messages) but on one user feed calls relatedMessages about 30 pages worth logging messages! It looks like it might be reading ALL the messages a given user has posted.

dominictarr:
Oh, also since we now we have flat threads and they all point to root we don't need to use `relatedMessages` we can just use `links({dest: msg, rel: 'root', source: '%', values: true})` and it will do that only once, instead of doing that for again for each message.

pfraze:
>  but on one user feed calls relatedMessages about 30 pages worth logging messages! It looks like it might be reading ALL the messages a given user has posted.

Ah good find, I'll investigate that.

> we can just use links

I need the votes on the replies.

pfraze:
Easy fix. Here's what's going on in the codebase.

There's an oversized component, [msg-list](https://github.com/ssbc/patchwork/blob/86452344896ffa0c41f1d908e06b08b1ac0e4b22/ui/com/msg-list/index.jsx), which has become a "workhorse." It's used very frequently, which is good, but it has way too much functionality and needs to be decomposed. (This was one thing my aborted refactor was meant to address.)

[This is the pipeline for messages](https://github.com/ssbc/patchwork/blob/86452344896ffa0c41f1d908e06b08b1ac0e4b22/ui/com/msg-list/index.jsx#L313-L321). Notice that filtering occurs after [`processMsg`](https://github.com/ssbc/patchwork/blob/86452344896ffa0c41f1d908e06b08b1ac0e4b22/ui/com/msg-list/index.jsx#L271-L277), which calls [`u.getPostThread`](https://github.com/ssbc/patchwork/blob/86452344896ffa0c41f1d908e06b08b1ac0e4b22/ui/lib/util.js#L306), which is the source of all the RPC calls you logged.

[...]

PR #76, opened 2015-11-07
fix: com/msg-list now filters before fetching thread data, to avoid needless RTs

pfraze:
See https://github.com/ssbc/patchwork/issues/74


pfraze:
@dominictarr let's merge this but leave #74 open

pfraze:
@dominictarr let's merge this but leave #74 open

Issue #65, opened 2015-11-05
rework: flag/unflag


dominictarr:
I created a new account, made a post, flagged it, then unflagged it.

But if I went back to the main feed this message was still shown as flagged.
If I restarted patchwork it is shown correctly, but not if I simply switch back to the feed view.

dominictarr:
with another flag added so it shows as flagged, I switched to my own account, and viewed the feed. Since the new account is followed by my pub I had it replicated but it was shown and not flagged. I think this surely should have been flagged. It was flagged by the account that posted the message, who was friend of a friend (via an autofollow).

My connection to that user as a flagger shouldalbe as close as my connection as a flagger... So theirfore it should accept the flag. Also, logged in as my self it didn't show anything about whether that was flagged.

pfraze:
> I created a new account, made a post, flagged it, then unflagged it.

> But if I went back to the main feed this message was still shown as flagged.
If I restarted patchwork it is shown correctly, but not if I simply switch back to the feed view.

That's a bug, will look into it.

> I think this surely should have been flagged. It was flagged by the account that posted the message, who was friend of a friend (via an autofollow).

Ok, I think that's the right behavior for now.

pfraze:
#81 and #82 together solve this.

PR #81, opened 2015-11-08
Mixmix Msg Chrome

pfraze:
This updates the message chrome to match @mixmix's mockup.

The mockup:
mix-mockup

This PR:
screen shot 2015-11-07 at 7 57 17 pm
screen shot 2015-11-07 at 8 07 09 pm

Not a 100% match, but close. I haven't changed how new replies are indicated yet.


dominictarr:
One thing i really liked about @mixmix's pitch for it is it has a rational for everything.
And the reasons he gave where clearly consistent and would be applied to other things.
eg. in left-right languages the future is to the right, so put actions on the right. group things with similar functionality.

It was really interesting because you can agree to the principles and immediately know what the correct thing is. In mix's mockup, the dotmenu is on the right, but in your image save is on the right, that feels intuitively wrong to me... but I'm not sure why. Is there a principle we can come up with that rationalizes why it should be one or the other?

pfraze:
Hmm. Good question. 

Putting the dropdown on the right continues the left-to-right flow, which does feel natural.

Putting the save-btn on the right gives it dominance, since top-right is a prominent position, which is good for a commonly-used item.

I think they look good either way.

mixmix:
Thanks for the feedback! I'm really flattered.

It's important to note I'm not a designer. This is as much a note to self
as to you guys. Every time i do some design work i think i'm doing ok, then
when i work with a designer I'm totally humbled by their experience and
skill.

I'll see if i can nab some designers at enspiral for a quick user test when
i get back to nz.

pfraze:
@mixmix sounds great

https://github.com/ssbc/patchwork/issues/65 got hit in this PR. https://github.com/ssbc/patchwork/commit/d8d877c09560cfcf128bdaa4d2755cd969607f61 (https://github.com/ssbc/patchwork/commit/d8d877c09560cfcf128bdaa4d2755cd969607f61#diff-082b1ce5b9da9f71152ff2aa459ba81aR137 specifically) should solve the first issue I quote in https://github.com/ssbc/patchwork/issues/65#issuecomment-154273650. Second issue still exists.

I'll go ahead and swap the search and dropdown btns, because there have been 3 votes for that and 1 for the other (I vote for both). Then I suggest we merge this, and can keep looking at it in the future.

pfraze:
screen shot 2015-11-08 at 7 40 33 pm

@dominictarr @mixmix ok to merge?

mixmix:
not sure what the search / dropdown thing is, but I've been living in `mix-chrome` for the last day and it's **comfy**!!

reckon it's good to merge (conflicts aside)

Issue #51, opened 2015-10-28
Security error prevents webcam loading


clehner:
When I click the webcam icon, the webcam media area is empty, and an error appears in the console:

> main.html:1
> Refused to load media from 'blob:file%3A///31acfe39-df1a-45d8-bfb7-92556a0a6b82' because it violates the following Content Security Policy directive: "default-src 'self' http://localhost:7777 http://localhost:7778". Note that 'media-src' was not explicitly set, so 'default-src' is used as a fallback.

When I then click the webcam media area, the following message is logged on the console, followed by the same error as before:
> /usr/local/src/patchwork/ui/lib/com/webcam-giffer-form.js:250
> Webm video encoded: 321 bytes

Debian testing amd64, Node v4.2.1, patchwork 4fb5789b5002bd587a273ef0a33bc2072b34baab

pfraze:
Thanks for the report, will look into it

dominictarr:
@clehner interesting. I get the same error. For me, _sometimes_ the webcam works, sometimes it doesn't. I'm also on (arch) linux.

PR #95, opened 2015-11-11
Update content security policy to allow webcam access (media-src blob:)

clehner:
This fixes the webcam problem I was having in #51

Also this adds an error handler to where the error would be caught without the fix (in case it fails for some other reason)


dominictarr:
@clehner so now this fully works for you?
@pfraze and I where discussing dropping this but if it can be made to work easily, maybe we should keep it?

clehner:
It works fully for me, yes. I can see maybe the feature is a little gimmicky, but I think it's nice

dominictarr:
It would be less gimmicky if it could also be use to take still photos.
I've wanted to do that occasionally.

pfraze:
Thanks for the fix, @clehner. I'll merge this once we know what the plan is with the webcam tools.

Issue #88, opened 2015-11-09
Show missing messages in threads


pfraze:
When a message in a thread replies to a message which is not locally available, the UI should indicate that there's an unknown post.

dominictarr:
have you seen any cases like this yet?

pfraze:
I dont think so, but I also dont think we'd know for sure w/o code to render the missing msg.

dominictarr:
We could collect all the conversations and traverse the internal links.
That might work.. Hmm unless you already have everything. What about we
just add logging for a missing message? Actually sometimes this might
happen during replication , depending on order.

pfraze:
Here's the scenario where I see this happening:

Alice `--follows-->` Bob `--follows-->` Carla `--follows-->` Dan

Alice's FoaF doeesnt reach Dan.

- Bob makes a post (msg 1)
- Dan replies to Bob's post (msg 2)
- Carla replies to Dan's reply (msg 3)
- Alice opens Bob's thread, and sees msgs 1 & 3. 2 is absent.

I've got some code almost ready to detect this and render a "this message is missing" card.

pfraze:
Done

PR #97, opened 2015-11-12
ui: threads now render 'missing message' cards when a reply refers to a branch that's not found

pfraze:
https://github.com/ssbc/patchwork/issues/88

When a message references a `branch` that's not in the `relatedMessages` results, this code will render the following:

screen shot 2015-11-12 at 12 57 32 pm

Clicking on the blue text expands into a longer explanation:

screen shot 2015-11-12 at 12 57 39 pm

### DIFF
--- a/ui/com/msg-thread.jsx
+++ b/ui/com/msg-thread.jsx
@@ -21,15 +21,30 @@ export default class Thread extends React.Component {
   // helper to do setup on thread-change
   constructState(thread) {
     // collapse thread into a flat message-list
-    var added = new Set()
+    let msgIds = new Set([thread.key])
+    let msgs = [thread]
+    ;(thread.related||[]).forEach(msg => {
+      // filter out duplicates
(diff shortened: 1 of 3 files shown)


pfraze:
Here's a thread with 2 replies with the same missing parent, and then a third reply with a different missing parent:

screen shot 2015-11-12 at 1 12 19 pm

pfraze:
Tweaked that help copy to say, "Some of the messages in this thread may reference it."

dominictarr:
:+1:

dominictarr:
It's weird to see that message, because you are telling the user something but not giving them any clues about what they can do about it. That doesn't empower the user.

It would probably be better to download that message, even if it's only shown with a pubkey as id, and not a name...(or something) maybe require a user click to request that message manually from a server...

But, I don't think that is a priority until someone complains about this.

pfraze:
Yeah I'm in agreement about all of that

Issue #55, opened 2015-11-03
Tracking relationships with Pubs


pfrazee:
I'd like to keep a registry of pubs that the user has joined. Why:

 - When the user publishes new data, we could prioritize syncing with the users' joined pubs, to ensure quick delivery.
 - If the users' joined pubs are not reachable, we can alert the user.
 - If the user hasn't joined any pubs yet, we can give instructions to do so.

The registry could be updated when an invite is successfully used.

@dominictarr you have any objections to this? Do you think this registry should live in the invites plugin, or in PW?

dominictarr:
isn't this as simple as filtering the followers by those who are pubs? This is something that would be very easy to calculate as a query.

mmckegg:
I think some of these issues are being addressed in the changes to gossip in the latest scuttlebot branches in git-ssb. There is also discussion about prioritizing connections to pubs you follow.

Closing!

PR #101, opened 2015-11-19
re-arrange connections layout

mixmix:
Hey @pfraze here's step in the direction I've designed.

The part I'm not sure about how to do is how to sort the peers. 
I need to know : 
 - how to distinguish a peer as an wifi connected user 
 - pubs vs. users  (I'm hoping there's a record of invites records somewhere?)

(At the moment the sort of just by friends)

selection_101

### DIFF
--- a/api/index.js
+++ b/api/index.js
@@ -135,11 +135,12 @@ exports.init = function (sbot, opts) {
   api.getIndexCounts = function (cb) {
     awaitSync(function () {
       cb(null, {
+        newsfeedUnread: state.newsfeed.countUntouched(),
         inbox: state.inbox.rows.length,
         inboxUnread: state.inbox.filter(function (row) { return !row.isread }).length,
         bookmarks: state.bookmarks.rows.length,
         bookmarksUnread: state.bookmarks.filter(function (row) { return !row.isread }).length,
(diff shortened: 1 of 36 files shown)


pfraze:
> how to distinguish a peer as an wifi connected user

In `/ui/lib/util.js` you can find the `getPubStats` function. It uses the `ip` module to check if addresses are on the LAN (`ip.isPrivate`). A lan peer will be `ip.isPrivate(addr) && !ip.isLoopback(addr)`.

> pubs vs. users (I'm hoping there's a record of invites records somewhere?)

It's a bit weird; I raised the issue previously in https://github.com/ssbc/patchwork/issues/55

Basically the current way to do it is, to see if the IP is public. You're a "member" of that pub if the user follows you.

---

Some thoughts:
 - Is "Connections" the best name for this section? What about "Network"?
 - Should the "Users" section be called the same thing as the icon at top? "Local", "Local Area", or "WiFi" maybe? (WiFi isnt 100% correct but I think most people use "WiFi" to describe the LAN nowadays).
 - What if the leftnav item showed the number of active Global and Local connections, maybe in small text underneath?

pfraze:
:guitar:

PR #102, opened 2015-11-20
Better blobs http server

pfraze:
1. Uses a cleaner http server implementation (originally made for https://github.com/ssbc/brickhouse)
2. Uses `ip` module `isLoopback` to check whether requests are from localhost
3. Uses `blobs.want` instead of `blobs.has` so that requests can stay open while the blob is downloaded from the network.


pfraze:
@dominictarr RFr

PR #105, opened 2015-11-21
add time-ago in words for times > 1 min

mixmix:
**Purpose** : making connection times easier to read. 

selection_102

Behaviour is to say "moments ago" instead of reporting every second change, which is quite distracting. 

I've dropped the "Last seen" prefix. Interested know if others find it confusing.
open to language changes


pfraze:
We've already got the `moment` dep for this, so I'll implement using that

PR #107, opened 2015-11-22
switch order of global + local

mixmix:
Dominic suggested switching the Local section to the top of the network Page.
I think he's right. 

He also suggested hiding it if there weren't any local connections. 
Tried this and don't think it's the best - if users only check there once, and that section isn't there, it will lead them to believe this is just a Pubs page.


pfraze:
:+1:

PR #133, opened 2015-11-28
Use profile pics in inbox views

pfraze:
Figured this might look nice:

screen shot 2015-11-28 at 12 48 05 pm

Yea/nay?

### DIFF
--- a/ui/com/index.jsx
+++ b/ui/com/index.jsx
@@ -49,6 +49,16 @@ export class UserPic extends React.Component {
   }
 }
 
+export class UserPics extends React.Component {
+  render() {
+    let ids = this.props.ids
+    ids.sort()
+    return <span>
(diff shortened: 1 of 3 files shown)


sedmonds:
Nice UI alternative, but just that. I'd love to see what happens when adopting the rule of silence in one styleset.

pfraze:
@sedmonds what would the rule of silence in a stylesheet look like?

sedmonds:
A lot. Will attempt to distill my thoughts on Patch.

mixmix:
I'd love to start a list of things to allow users to customise. This seems like a nice thing to offer as toggleable
Personally I love the images, adds color and fast to read. Might be phishing-vulnerable.

dominictarr:
I would approve of this - but @mixmix does have a point about phishing - it should be possible to set someone else's profile before we merge this.

Issue #120, opened 2015-11-24
Topics: message format & basic behaviors


pfraze:
(This issue is just about the data model for public groups.)

Groups are effectively just a filter on post messages.

 1. How do posts set their group? Can posts have multiple groups?
 2. How are groups declared? Are they unique entities, declared by a message, which the member-messages link to? Or are they simply text-labels?

pfraze:
One thing to decide is, are we implementing Groups, or Topics?

Topics are just a label (sort of like hash-tags) for categorizing posts.

Groups usually have members. In a public group, membership wouldnt do much more than be a social-signal. In secret groups membership will matter more.

If groups are things which must be declared as messages, and linked-to, then you need the original message to know what the group is.

pfraze:
I suggest we just do this as topics, with the following schema:

```js
{
  type: 'post',
  topic: String|[String],
  ...
}
```

Each topic is a string. You can have multiple topics to cross-post in multiple groups. Patchwork builds an index for each topic.

Two options on how to handle topic comparisons:
 1. Require an exact match to consider two topics to be ===.
 2. Make topics case-insensitive, and strip non-alphanumeric characters, such that `'foobar' === 'Foo Bar' === 'foo-bar' === 'foo_bar' === 'foo$bar'`. Patchwork's UI will forbid spaces and non-alphanum characters, to discourage their use. This is to try to reduce the possible variations on the same intended topic.

At first I was inclined toward 2, but maybe it's better to do 1? Why: 1 doesnt presume everybody speaks english, and it gives more flexibility to users, even if can result in more mis-posting.

pfraze:
Some composer mockups:
![screen shot 2015-11-25 at 3 29 41 pm](https://cloud.githubusercontent.com/assets/1270099/11409982/6d1de11c-9389-11e5-94d0-441dada0fb7d.png)
![screen shot 2015-11-25 at 3 29 52 pm](https://cloud.githubusercontent.com/assets/1270099/11409981/6d1c8204-9389-11e5-93ad-8d6e59c5786d.png)

pfraze:
Making topics more distinct in the composer:
![screen shot 2015-11-25 at 3 38 16 pm](https://cloud.githubusercontent.com/assets/1270099/11410132/aa2a73d0-938a-11e5-9c6d-7c12788fb224.png)

pfraze:
Mockup of topics in the sidenav:
[...]

PR #134, opened 2015-11-28
Topics

pfraze:
Add topics, per https://github.com/ssbc/patchwork/issues/120


pfraze:
screen shot 2015-11-28 at 4 36 23 pm
screen shot 2015-11-28 at 4 36 32 pm

pfraze:
This PR relies on https://github.com/ssbc/ssb-msg-schemas/pull/12, so it wont work without that

It adds the Topic post-type in the composer. The post is categorized under the topic you choose. (An autocomplete helps you choose one already in use.)

@dominictarr @mixmix @aredridel et al, thoughts?

pfraze:
This is a really low-impact change. The only effect it has on the dataset is, it adds a `topic` field to posts. If we decided to roll this back, we'd ignore the `topic` values on posts, and theyd merge back into the main newsfeed. No biggie!

I could fix the PR so it doesnt depend on ssb-msg-schemas being updated, and then we could merge, if we want to just give it a shot.

pfraze:
Ah, something funny - people who dont have this feature will not be aware of the topics. These posts will just show up in their main feed. Not a big deal, but... interesting.

substack:
How do the topics scale for browsing in the UI when there are a few dozen, or few hundred of them? Do the topics show up with notification counts like in irc clients?

pfraze:
> How do the topics scale for browsing in the UI when there are a few dozen, or few hundred of them?

We'll need a way to:
 1. pin the topics you want to see always, and
 2. discover the topics you havent pinned that are getting the most activity

I was planning to add a  "more..." button to the right end of the nav, which would take you to a topics page. For overflow of pinned topics, it might be like how Chrome does its bookmarks -- it becomes a dropdown.

Alternatively, we could move the topics nav into the lefthand nav, which can scroll vertically as far as we like. I'm wholly open to this change (we might need to do it!).

> Do the topics show up with notification counts like in irc clients?

Open question. I figured I'd show an unread count on them, that auto-updates.

pfraze:
I just updated this to work without the ssb-msg-schemas PR being merged. If you want to try it, feel free to checkout this branch (`topics`)

pfraze:
is it a good idea to add topics to `type: post`, like this?

why it not might be:

 1. semantics: `type: post` is a fairly generic core type, for publishing to the feed. Does adding topics fit the premise of the type? Is it better to create a new type - `type: topical-post` ?
 2. effects: what happens if topics arent supported by a client in the network? (at the moment, that's everybody but me.) suppose I, seeing the topic-filtering, decided to post 10 cat gifs to `topic: "catgifs"`. To me, it's an appropriate action, because Im posting to a niche-space. To everybody without topics, it's major main-feed spam, because there's no filter. So I get flagged for spam, I lose all my friends, and I die lonely.

aredridel:
Good points, @pfraze

ahdinosaur:
> is it a good idea to add topics to type: post, like this?

i prefer the simplicity of a single post type that links to other messages through a common mechanism (links and backlinks). to expand: if we imagine a topic as an immutable identifier (#catgifs), then a topic post is a message that references a topic identifier and a topic channel is a list of all messages that backlink to that topic identifier. maybe this is crazy, but i'm still hooked.

pfraze:
Proposal: main feed shows messages in *all* topics, and the topic-views filter down.

pfraze:
@dominictarr any comment on this PR?

pfraze:
(the UI needs work, but the basic concepts should be discussed)

dominictarr:
There is a good question here about how do we manage changes. And how can we design patchwork to accommidate changes. For example, the feed etc are basically hard coded - but potentially, it could be based on a filter that was customizable?
We could make a new type of message, but then that is more hard coding, especially since it's basically the same as a post. (it has root and branch and mentions, and a markdown text field)

Maybe we should use duck typing and deprivledge the type field?

pfraze:
What constitutes a breaking change to the `post` type? I'd argue:

 - As in the PR currently, it is a breaking change (https://github.com/ssbc/patchwork/pull/134#issuecomment-160442126 describes why)
 - https://github.com/ssbc/patchwork/pull/134#issuecomment-160835968 describes a version that is non-breaking (accidental spamming would not occur)

Why -- it changes the shared "contract" for what's put in the main feed:
 - Before PR: if a `post` is made without `recps`, it's in the main feed.
 - After PR: if a `post` is made without `recps` or `topic`, it's in the main feed.
 - With https://github.com/ssbc/patchwork/pull/134#issuecomment-160835968: behaves as Before PR.

Per your thinking @dominictarr, one idea proposed while discussing with @jrf0110 was, strict defensive duck typing: ignore any message which has unknown fields. Earlier versions would therefore filter out these new `topic` posts (but after upgrading the messages would appear). The reason being, a misunderstanding could have bad consequences, so it's better not to accept the message.

I dont think hardcoding a new post-type is such a big deal, but I agree it's not very elegant.

ahdinosaur:
relevant: [traits](https://en.wikipedia.org/wiki/Trait_%28computer_programming%29)

dominictarr:
I think I much prefer the proposal that the main feed is ALL posts, and then other views are filters.
Then we just have a "general" topic that is `!msg.topic`.
The main feed is gonna get really noisy as more people arrive anyway...

jrf0110:
I think it's important distinguish whether or not topics is a separate app from the main feed. For instance, if Patchwork were to create Event Organization functionality, it would be clear that a new post type is required; `type: 'event'`.

However, what about conversations inside the event? It would be tempting to add yet another field to the `post` type (something like `event_id` or whatever), and rely on duck-typing to know that this post is only shown in the context of an event. However, we can see that this only takes us so far scale-wise. It creates a mental dependency on all other types of posts as you question whether or not the field you're adding conflicts with something else.

My proposal is to keep the basic `post` type as simple as possible. This is the fundamental unit of PW. Any extensions to this functionality should be done through a pluggable infrastructure. Thinking this way, the Events application would probably introduce two types: `event`, `event-post` - where `event-post` is a `post` inside an event. It has a rigidly defined and versioned schema.

--

> I think I much prefer the proposal that the main feed is ALL posts

The main feed need not be all of `type: 'post'`, but rather, it should be an aggregate of all post types. PW should use software to show the user what's pertinent.

PR #147, opened 2015-12-02
Ngraph friends

mixmix:
selection_128

spin it up, it's dynamic! 

would be good to talk about where this should live. 
Also, worth noting this thing spins up the CPUs a bit because it's iterating hard ...


dominictarr:
It looks cool, but what is it useful for?

mixmix:
Good question. Is it nice enough 'sizzle'  (or could it be)  to want to
include?

I l'm a bit stalled out on getting mouse over labels.
I think this doesn't help discovery, and I'm not sure it helps people
understand connections / the network.

dominictarr:
I could imagine it being a really cool animation in some context... maybe showing it populate as you replicate stuff on initial sync?

mixmix:
Oo, nice idea... Maybe showing the network be built...

pfraze:
I like that idea. We could also put it into a little network dashboard, along with stats like the scripts you recently made @dominictarr, and stats about your reach. "sizzle" / "mildly interesting data"

ui/views/sync.jsx
```diff
@@ -28,6 +33,212 @@ function isNotLAN (peer) {
...
+      'stroke-width': linkStrokeWidth,
+      'stroke-dasharray': linkStrokeDasharray,
+      opacity: linkOpacity
+    })
+  })
+
+  let group = ngraphSvg.svg("g", {transform: "translate(10 20)"})
+  let text = ngraphSvg.svg("text", {fill: 'black'}).text("John Dot Awesome") 
+  let img = ngraphSvg.svg("image", {
+    //'xlink:href': 'http://localhost:7777/&xISx10wRtEirc1+LEHp0khJc5lrsybHBnIry78pan9U=.sha256?fallback=img&name=JohnDotAwesome.png',
```
mixmix: @pfraze I'm having a heaps of trouble getting an image to render in SVG inside Patchwork.

Note the same pattern works in JSFiddle http://jsfiddle.net/4s6mucuq/
(assuming blob names are the same .... )

you can see on link 169 / 170 I've tried changing these attributes and this comes through in electron, but then displays nothing ... 

if you uncomment this line you get an error about invalid namespace. Not sure what's needed there but have tried a few things. ngraphSvg.svg is actually just this file, and setting attributes uses this function which is called setAttributeNS  https://github.com/anvaka/simplesvg/blob/master/index.js#L81

---

if you could spin it up and confirm I'm not crazy or missing something dumb like z-index that would be rad.
Further to that, any suggestions ?

pfraze: I'm doing client work atm, but quick check:

See if the version of https://www.npmjs.com/package/pull-identify-filetype is `1.0.1`. If it's not, update and restart. I recently fixed svg hosting in PW with its last patch

mixmix: Alas that's not it
Appreciate you taking the time to post. Can catch up on this whenever

pfraze: I cant figure this out!

mixmix: Is resolved now, spin the branch up!

pfraze:
:+1: :+1: :+1: will code review soon

mixmix:
There's a small fix needs doing about identifying local connections as
relayPeers...  Which isn't super accurate. I think i might be able to show
live connections too

ui/lib/util.js
```diff
@@ -111,3 +111,18 @@ exports.getPubStats = function (peers) {
     hasSyncIssue: (!membersof || (!membersofUntried && !membersofActive))
   }
 }
+
+exports.getRelayPeerIds = function (peers) {
```
pfraze: We're still using the term Pubs, and I think that'll be a shorthand for Public Peers.

pfraze: Is this a function to get pubs, or just peers you connect to? If the latter, this ought to be more like... `getConnectedPeerIds`

pfraze:
Ok that's my only note. Overall, code looks really good. Going to pull and try it now...

pfraze:
Looking great. Something funny and pretty appropriate: if there are any unconnected nodes, they fly off in the beginning.

PR #154, opened 2015-12-04
support rotation of avatars while cropping

dominictarr:
My new avatar was showing sideways so I added rotation to the cropping tool.

(I don't know why, because it does render up the right way when you open it in any other photo app, but anyway, now we can rotate avatars)

### DIFF
--- a/ui/com/image-input.jsx
+++ b/ui/com/image-input.jsx
@@ -37,14 +37,9 @@ export default class ImageInput extends React.Component {
   onRotate(e) {
     e.preventDefault()
     e.stopPropagation()
-    this.setState({ rotation: (this.state.rotation + 1) % 4})
-
-    //MAGIC setTimeout. without this, it draws the last state not the current state.
-    //it wouldn't show correctly until you fiddled with another control.
-    //I couldn't figure out why, but this fixes it...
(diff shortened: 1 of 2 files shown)


pfraze:
ok cool, let me poke at it and see if I cant get rid of that settimeout magic

dominictarr:
awesome!

dominictarr:
did you figure this out? can we merge?

pfraze:
I'll try to get this tonight

pfraze:
@dominictarr setState() is an async operation, you just needed to use a callback

dominictarr:
oh i didn't realize it was async because the other uses didn't use it with a callback.
Why is it async? does it update the state in the next frame?

pfraze:
From docs:

> setState() does not immediately mutate this.state but creates a pending state transition. Accessing this.state after calling this method can potentially return the existing value.

You only need to pass a CB if the following code depends on the updated `this.state`

dominictarr:
@pfraze another question: is there a way to have something trigger after the state changes? does it need to be in the render function for that? If you have something like a canvas in there (as in this code) it breaks the functional thing to explicitly call drawCanvas, after each update, really react should just call that somehow.

PR #164, opened 2015-12-07
Inline composer

pfraze:
Removes the floating action button, and puts the composer into the newsfeed and inbox UIs.

A future PR should add the composer to the profile page.


pfraze:
composer-expand

PR #165, opened 2015-12-09
New UI shell (move from leftnav to topnav)

pfraze:
This PR updates the overall UX. It moves nav from the left to a toolbar on the top, adds toggleable right columns, and increases the font-size of content.

screen shot 2015-12-08 at 8 50 36 pm
screen shot 2015-12-08 at 8 51 13 pm
screen shot 2015-12-08 at 8 51 22 pm


dominictarr:
:+1:  for toggleable collumns!

pfraze:
@mixmix I ended up moving the bookmarks into a righthand column, because of the inline composer. The feed view has the public composer, and the inbox view has the private composer... so it felt a bit weird to then have the bookmarks page, with no composer. And, the 4 sections now make sense to me (public view, private view, contacts, network). I felt bookmarks really made more sense as a navigational tool.

I know that wasnt your preference, so I'll remain open to discussing it!

PR #171, opened 2015-12-10
Fix peer dependency errors

jrf0110:


jrf0110:
Oh, and I guess update some other unrelated deps. If you're into that sort of thing

pfraze:
Ahhh thanks yo

jrf0110:
image

PR #177, opened 2015-12-12
Ngraph friends

mixmix:
hey @pfraze I've done a little fix, and also added dynamic display of active connections. 
Looks like this : 

anim

I've found a way to add clickable links to the svg so you can browse to a user profile, but I think the solution I tired fails to unload the listeners on the SVG. 
Probably a problem for later / not at all

### DIFF
--- a/ui/views/sync.jsx
+++ b/ui/views/sync.jsx
@@ -57,7 +57,7 @@ class PeerGraph extends React.Component {
   }
 
   componentWillReceiveProps (nextProps) {
-    this.state.graph = updateGraph(this.state.graph, nextProps.peersForGraph, nextProps.contactedPeerIds)
+    this.state.graph = updateGraph(this.state, nextProps.peersForGraph, nextProps.contactedPeerIds)
   }
 
   componentWillUnmount () {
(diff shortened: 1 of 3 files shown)


ui/views/sync.jsx
```diff
@@ -131,32 +134,41 @@ function createRenderer (graph, el) {
...
       r: radius,
       cx: 2*radius,
       cy: 2*radius,
       class: nodeClass.join(' '),
-      peerId: node.data.id,
+      id: node.data.id,
       text: node.data ? node.data.name : undefined
     })
+    const link = ngraphSvg.svg("a")
+    //const link = ngraphSvg.svg("a", { link: '#/profile/'+encodeURIComponent(node.data.id) })
```
mixmix: here's the line that adds a non-react router link and throws an issue when you click on links 
says: 

```
TypeError: Cannot set property 'innerHTML' of null
    at SVGSVGElement.<anonymous> (/home/mix/projects/patchwork/ui/views/sync.jsx:86:55)
```

pfraze:
:+1: I'll look into the error. Will merge later today!

pfraze:
@mixmix hey, suggest you start using new branch names after a PR gets merged, just to avoid confusion with name-clashes

pfraze:
This thing is really fun to watch :)

mixmix:
yeah agree. this was a case of wasn't quite done when you previously merged. Which was fine .. it was done-enough. 

I wouldn't sweat it on that error. If you know how to get around it that would be sweet, otherwise there are more important things

pfraze:
@mixmix yeah https://github.com/ssbc/patchwork/commit/bbaebb0be87f40e5f7b83e1df8a04fc2d657d070 fixed it, merging now

Issue #170, opened 2015-12-10
Add context menu


pfraze:
Right clicking needs a context menu, especially for images and links!

pfraze:
Big thanks to @clehner

PR #186, opened 2015-12-13
Context menu

clehner:
Here is a context menu. It includes up to four things:
- Open Link
- Copy Link Location
- Save Image As
- Copy Image Location

I think it might be better to copy message IDs instead of localhost link/image URLs where possible but I haven't implemented that yet.

### DIFF
--- a/app/menu.js
+++ b/app/menu.js
@@ -1,7 +1,7 @@
 var Menu = require('menu')
 var dialog = require('dialog')
 
-var isMac = (process.platform == 'darwin');
+var isMac = (process.platform == 'darwin')
 
 module.exports = function (window) {
   var template = [
--- a/ui/contextmenu.js
+++ b/ui/contextmenu.js
@@ -1,0 +1,116 @@
+var remote = require('remote')
+var Menu = remote.require('menu')
+var MenuItem = remote.require('menu-item')
+var dialog = remote.require('dialog')
+var clipboard = require('clipboard')
+var app = require('./lib/app')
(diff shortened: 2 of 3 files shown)


pfraze:
:+1: you :guitar: will review and merge tomorrow

pfraze:
This is a great start. It needs to have a little more intelligence about the kinds of objects and links in Patchwork.

 1. Open Link: a-ok
 2. Copy Link Location: if the link is to a message, user, or blob, it should get their SSB ID instead of the HTTP url. If the link is to a website, it should get the HTTP URL.
 3. Copy Image Location: same story.
 4. Save Image As: needs to set a good default name on the save-as dialog, because the hash is not any good. The best way to do this would be, to use the `alt` attribute on the image, and make sure the alt is always populated across the UI. (It currently is populated for inline images on a post, but not for profile pics.)

pfraze:
@clehner if you wont have time to make those changes, let me know and I'll do them.

clehner:
@pfraze that makes sense. I'll update it to do those things

clehner:
"Save Image As" now gets the filename from the name suggested in the URL if it exists, which works for inline attachments and profile pics, and falls back to `alt` otherwise.
I also added some text editing items following the behavior of web browsers and text editors.
That's about all I have in mind to add right now

pfraze:
Great, thank you for the contribution!

pfraze:
closes https://github.com/ssbc/patchwork/issues/170

PR #187, opened 2015-12-13
Make menu bar cross-platform friendly

clehner:
This PR changes menu item things to work better on Linux (and maybe BSD or Windows too).

- Use idiomatic accelerators (CmdOrCtrl)
- Make Aqua-specific items conditional
- Use role instead of selector. I don't think this should cause breakage on OS X since its what the [docs](https://github.com/atom/electron/blob/master/docs/api/menu.md) use, but I haven't tested this.
- Implement Quit item

The About menu item still causes an error "Menu.sendActionToFirstResponder is not a function" but there is no UI code to link it to so I didn't change it.

### DIFF
--- a/app/menu.js
+++ b/app/menu.js
@@ -1,6 +1,8 @@
 var Menu = require('menu')
 var dialog = require('dialog')
 
+var isMac = (process.platform == 'darwin');
+
 module.exports = function (window) {
   var template = [
     {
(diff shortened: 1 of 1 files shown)


pfraze:
:+1: will evaluate soon, along with the context-menu PR.

pfraze:
thanks!

PR #178, opened 2015-12-12
New topics

pfraze:
Decided to supplant https://github.com/ssbc/patchwork/pull/134 with a new PR, that picked some of the working code out and implemented a new UI

PR #189, opened 2015-12-13
Channels

pfraze:
Replaces https://github.com/ssbc/patchwork/pull/178, same functionality as that PR but called channels instead of topics (including in the msg schema)


pfraze:
@dominictarr you ok with the channel change? This updates the post message schemas (from `.topic` to `.channel`)

dominictarr:
@pfraze but where is the pr for ssb-msg-schemas?

pfraze:
Oh, not yet done. The change will be to add a `channel: String` to the `type: post`. I just manually set that attribute in this PR.

dominictarr:
if it's not easier to make the pr to that module then we are doing something wrong

pfraze:
catch-the-bubbles
ok gimme a sec

pfraze:
https://github.com/ssbc/ssb-msg-schemas/pull/13

PR #194, opened 2015-12-15
Thread view panel

pfraze:
This PR does some styling update to the messages, and adds a toggleable message-view on the right nav, which  shows threads you click. If you disable the thread-view panel, it'll do a normal page navigation.

**Updated feed styles**
screen shot 2015-12-15 at 1 22 12 am
**Updated inbox styles**
screen shot 2015-12-15 at 1 22 51 am
**thread selected**
screen shot 2015-12-15 at 1 23 12 am
**in the feed view**
screen shot 2015-12-15 at 1 23 19 am

### DIFF
--- a/ui/com/msg-list.jsx
+++ b/ui/com/msg-list.jsx
@@ -45,8 +45,10 @@ export default class MsgList extends React.Component {
     // handlers
     this.handlers = {
       onSelect: msg => {
-        app.emit('open:msg', msg.key)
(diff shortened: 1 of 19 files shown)


pfraze:
Preview pane is now configurable within the newsfeed's config btn.
Even if we decide against the preview panel, there are style updates in here that should happen! So, merging.

pfraze:
Got some negative early feedback, so this now defaults to being off. If you want to test it, turn it on in the newsfeed config.

PR #201, opened 2015-12-18
Persist zoom

clehner:
This lets the user's zoom level be persisted across application launches


pfraze:
:+1: solid. We're going to need to organize our localstorage use someday but this is great for now

PR #202, opened 2015-12-18
Menu updates

clehner:
General changes
- Add a "Go" menu for navigation items. Move Reload into it.
  - Add back/forward
  - Add items for navigating to top-level UI (Feed, Inbox, etc)
  - Add item for jumping to the Search bar
- Add items for toggling the UI nav right sidebars

Non-OS X changes
- Remove About item which doesn't work. We can add it back after we make our own About window with useful info like the license, list of contributors, link to web site, etc
- Remove Minimize item. Generally window managers if they support Minimize provide their own key/interface for it. (Correct me if yours doesn't and you think this item should stay)

### DIFF
--- a/app/menu.js
+++ b/app/menu.js
@@ -220,15 +220,14 @@ module.exports = function (window) {
         }
       ]
     },
-    {
+    isMac ? {
       label: 'Window',
       submenu: [
         {
           label: 'Minimize',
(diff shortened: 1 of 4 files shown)


pfraze:
:+1: excellent update, @dominictarr and @substack you'll be happy to see the shortcuts added here.

PR #204, opened 2015-12-19
Inline Find

clehner:
First crack at a inline find bar (Edit menu -> Find)

- The searching only finds text within individual text nodes so it won't find text if it is broken up by formatting. To fix this would make the algorithm more complicated but could be added in the future.
- Sometimes it finds text that isn't visible because it is hidden in text-overflow or is underneath the search bar itself. I'm not sure what to do about this.
- I've only lightly tested this so there may be other bugs.
- The UI could use a little help.
- Highlighting results isn't implemented but could be added later.

Edit: The search behavior is sufficiently reusable/abstract that I made a [separate module for it](https://github.com/clehner/text-node-searcher).

### DIFF
--- a/app/menu.js
+++ b/app/menu.js
@@ -83,10 +83,24 @@ module.exports = function (window) {
           type: 'separator'
         },
         {
-          label: 'Find',
+          label: 'Find...',
           accelerator: 'CmdOrCtrl+F',
           click: function () {
             window.rpc.triggerFind()
+          }
+        },
+        {
+          label: 'Find Previous',
+          accelerator: 'CmdOrCtrl+Shift+G',
+          click: function () {
+            window.rpc.findPrevious()
+          }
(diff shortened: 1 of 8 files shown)


pfraze:
This is off to a solid start. It needs some more styling work (though I like your basic choices here) and some usability tweaks (sometimes the highlighted text is hidden under the search bar). We can merge now and continue to work on it, or you can keep working on it as a branch, your call

clehner:
@pfraze I will improve the UI and try to fix the hidden text issues and then consider it good to merge

clehner:
Fixed the UI and scrolling issues (that I could find). Also added highlighting, and Find Next/Previous menu items. It should be good to go now unless there are other bugs.

pfraze:
:+1: another great patch

Issue #163, opened 2015-12-07
When hovering over a link, show the URL somewhere in the UI


mixmix:
url preview for links please...  make it work just like a browser

as mentioned in this private conversation with @pfraze  : 
%EuStmweGPLn4pRdUOAsAPBStzYpTOdNpxeivmHWkh/o=.sha256

hmm .. http://localhost:7777/%EuStmweGPLn4pRdUOAsAPBStzYpTOdNpxeivmHWkh/o=.sha256
guess that's blob-address .. but daydreaming ...

mixmix:
hah .. [http://localhost:7777/%EuStmweGPLn4pRdUOAsAPBStzYpTOdNpxeivmHWkh/o=.sha256](http://localhost:7777/%EuStmweGPLn4pRdUOAsAPBStzYpTOdNpxeivmHWkh/o=.sha256)

manually linking ...

mixmix:
wow I'm a derpster @ahdinosaur to the rescue again  !

pfraze:
?

mixmix:
Lord, i woke up at 5am that day and made a lot of mistakes.
Paying and closing wrong issue for example

pfraze:
Haha

ahdinosaur:
superseded by https://github.com/ssbc/patchwork/issues/520

PR #206, opened 2015-12-20
Tooltips for links in messages

pfraze:
This is a proposed solution to https://github.com/ssbc/patchwork/issues/163
When hovering a link inside a message, you'll get a tooltip that shows the target.

I'm unsure if we want tooltips like this (displaying next to content) or a status bar, like browsers do (displaying in bottom left). This uses [tether-drop](http://github.hubspot.com/drop/) which is quite good.

Screenshots:

screen shot 2015-12-19 at 6 26 23 pm
screen shot 2015-12-19 at 6 26 38 pm
screen shot 2015-12-19 at 6 26 45 pm
screen shot 2015-12-19 at 6 26 52 pm


pfraze:
In the future, we'll want hover-info (whether a tooltip or status bar) for ssb objects to be more informative: usercards, message summaries, file metadata, etc

dominictarr:
:+1: this would be useful to me.

Although I think I might prefer a status bar.

pfraze:
yes, Im torn on that. Interested to hear the consensus opinion

PR #210, opened 2015-12-21
bump electron-prebuilt

wanderer:
This fixes an error on node v5. The newer versions electron-prebuilt have electron that is compiled against node v5.

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,6 @@
   "devDependencies": {
     "electron-builder": "^2.0.0",
     "electron-packager": "^4.1.3",
-    "electron-prebuilt": "^0.36.1",
     "less": "~1.7.5",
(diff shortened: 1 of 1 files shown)


pfraze:
:+1: this works! It required that, along with upgrading to node 5, I used npm v3.5.2. Going to test with node 4 real quick.

pfraze:
Also, I got some electron-api deprecation notices. Note to self: after merge, create issues to update API usage.

pfraze:
This seems to break compatibility with node v4. We'll need to let everyone know.

pfraze:
I'm going to defer fixing the electron compat changes for now, since they're only warnings. Related blogpost: http://blog.atom.io/2015/11/17/electron-api-changes.html

PR #211, opened 2015-12-21
Sqlite Experiment, Tracking PR

pfraze:
Patchwork currently uses leveldb and in-memory structures to control the dataset. This works, but it doesnt include a general query engine, and so new queries are tedious to write.

This branch is an experiment with using Sqlite to manage Patchwork's dataset. It is going to evaluate Sqlite for performance, flexibility, and space-efficiency, as compared to the leveldb+memory approach.


pfraze:
Current status: Demoed basic implementations of Patchwork's newsfeed, channel, and inbox indexes, using a sqlite in-memory instance. They work as expected.

Next: will need to fully implement all indexes and benchmark performance.

pfraze:
Good news / Bad news

The bad: loading the log into the database is slow. To get through my entire backlog takes roughly 30-45 seconds. This cost will only be incurred when playing through the entire backlog, which won't be often, but it's still an order of magnitude more than I'm comfortable with. If I can't reduce the load time now, I'm going to pause (and possibly abort) this experiment.

The good: query performance is very close to leveldb. Here are my functions:

```js
api.getThreadLeveldb = function (key, cb) {
    var done = multicb({ pluck: 1, spread: true })
    var msgCb = done()
    sbot.get(key, (err, msg) => msgCb(null, msg)) // suppress error caused by not found, will be reflected by missing `value`
    pull(sbot.links({ dest: key, keys: true, values: true }), pull.collect(done()))
    done((err, value, related) => {
      if (err) return cb(err)
      cb(null, { key: key, value: value, related: related })
    })
  }
  
  api.getThreadSqlite = function (key, cb) {
    state.sqldb.all('SELECT msg_key FROM post_msgs WHERE root_msg_key = ?;', [key], function (err, repliesMeta) {
      if (err) return cb(err)
      var done = multicb({ pluck: 1 })
      sbot.get(key, done())
      repliesMeta.forEach(function (rm) {
        sbot.get(rm.msg_key, done())
      })
      done(function (err, msgs) {
        if (err) return cb(err)
        cb(null, { key: key, value: msgs[0], related: msgs.slice(1) })
      })
    })
  }

  api.getThreadsLeveldb = function (cb) {
    var done = multicb({ pluck: 1 })
    state.newsfeed.rows.slice(-50).forEach(function (row) {
      api.getThreadLeveldb(row.key, done())
    })
    done(cb)
  }

  api.getThreadsSqlite = function (cb) {
    // query newsfeed
    state.sqldb.all(
      'SELECT root.msg_key AS msg_key FROM post_msgs AS root'
        +' LEFT OUTER JOIN msg_links AS root_recps ON root_recps.src_key = root.msg_key AND root_recps.rel = "recps"'
        +' WHERE root.root_msg_key IS NULL'
          +' AND root_recps.dst_key IS NULL'
        +' LIMIT 50;'
      , function (err, rootMsgs) {
      if (err) return cb(err)

      // fetch threads
      var done = multicb({ pluck: 1 })
      rootMsgs.forEach(function (row) {
        console.log('fetching', row)
        api.getThreadSqlite(row.msg_key, done())
      })
      done(cb)
    })
  }
```

and here are my results:

```
50x getThreadLeveldb
real  0m17.232s
user  0m14.187s
sys 0m2.046s

50x getThreadSqlite
real  0m17.036s
user  0m14.635s
sys 0m1.879s

1x getThreadsLeveldb
real  0m0.395s
user  0m0.304s
sys 0m0.042s

20x getThreadsLeveldb
real  0m8.740s
user  0m6.700s
sys 0m0.932s

1x getThreadsSqlite
real  0m0.560s
user  0m0.323s
sys 0m0.048s

20x getThreadsSqlite
real  0m10.484s
user  0m6.259s
sys 0m0.855s
```

pfraze:
Better news: turns out the default behavior for Sqlite is to fsync after every insert. However, a transaction will hold off on the fsync until committed, so wrapping the history replay in a single transaction got me down to 16 seconds. Let's see if I can do any better...

pfraze:
According to https://sqlite.org/speed.html, it's possible to do 25k inserts in less than a second. It appears I'm doing 38,955 inserts, which, even accounting for for differences in our data model, shouldn't be an order-of-magnitude higher. There's more to be gained here...

pfraze:
Down to 11 seconds if I let some queries run in parallel...

pfraze:
It's not clear yet whether the parallelization maintains correctness, so I'll use serial only for now, which is still at 15s for 36k inserts + 2 updates (to be more specific). [This guy got all the way to 72k/second](https://stackoverflow.com/questions/1711631/improve-insert-per-second-performance-of-sqlite) so clearly there's more to be eeked out here..

pfraze:
With `PRAGMA synchronous = OFF` and `PRAGMA journal_mode = MEMORY`, both of which make things a little riskier, I get 13.686s. Not a big update.

Note, it is possible the remaining overhead is nodejs vs C... but a whole 10x?

pfraze:
Ok, I'm having trouble eeking out much more. Performance does seem to drop with each query run, so it's not like there's one item (that I can find) responsible for all the overhead.

For reference, if I parallelize everything except the table creation, the benches I get are:
 - in-memory db: 6.146s
 - on-disk db: 8.191s

Which means we can get a 2x boost just by using more parallel queries, and I bet it'd be even faster if table creation didnt have to be serial.

PR #225, opened 2015-12-25
Update electron packaging

aredridel:


pfraze:
Hah, I keep forgetting this. Thanks!

PR #226, opened 2015-12-29
Last heard from

mixmix:
would love a review of this (I don't know what good patterns look like and would love to learn)

selection_201

known gotcha here is that I'm trusting the timestamp on the message, so we might hit that future error. I wonder if there could / should be methods for accessing timestamps which deal with this .... or what


ui/com/user-info.jsx
```diff
@@ -25,8 +25,8 @@ const FLAG_DROPDOWN = [
 class AutoRefreshingComponent extends React.Component {
   constructor(props) {
     super(props)
-    this.state = this.computeState()
-    this.refreshState = props => { this.setState(this.computeState(props)) }
+    this.computeState(null, this.setStateCallback.bind(this))
```
pfraze: Instead of creating a `setStateCallback` method on the class, I'd suggest just using an inline function.

```js
this.computeState(null, (err, state) => {
  if (err) throw err
  this.setState(state)
})
```

pfraze: Oh-- seeing that you have to use the CB twice, I'd just put it within the function. Just seems weird to be a class method to me.

```js
const handleCB = (err, state) => {
  if (err) throw err
  this.setState(state)
}
this.computeState(null, handleCB)

File: ui/com/user-info.jsx [hunk 2]
```diff
@@ -173,7 +189,7 @@ function sortFollowedFirst (a, b) {
 }
 
 export class UserInfoFollowers extends AutoRefreshingComponent {
-  computeState(props) {
+  computeState(props, callback) {
```
pfraze: Make sure all of the `computeState` methods use `callback()` instead of returning the state. Most of them dont.

dominictarr:
secure-scuttlebutt does track time the latest message was received, if https://github.com/ssbc/secure-scuttlebutt/blob/master/index.js#L330 also returned the latest timestamp. That is stored in lastdb like this: https://github.com/ssbc/secure-scuttlebutt/blob/master/index.js#L96

you can also get it directly from https://github.com/ssbc/secure-scuttlebutt/blob/master/index.js#L320-L322
I am happy to adapt secure-scuttlebutt to make this feature easier.

ui/com/user-info.jsx
mixmix: @pfraze are you suggesting setting that const inside the class definition like : 

```javascript
class AutoRefreshingComponent extends React.Component {
  const handleCB = (err, state) => {
    if (err) throw err
    this.setState(state)
  }
  constructor(props) {
    super(props)
    this.computeState(null, handleCB)
    this.refreshState = props => { this.computeState(props, handleCB) }
  }
...
```
?

I don't know of a way to do that, and pushing the const outside the class means you don't have access to `this` any more. 
I assume I'm missing some basic knowledge D:

mixmix:
@dominictarr I don't fully understand how Patchwork is tapping into ssb (I've had a read of patchwork  but haven't learnt about what muxrpc / ipc , so feel kinda fuzzy on how I would access stuff)

I feel like you making this stuff easier would make for a nice refactor once this is done? unless what you're proposing is gonna be synchronous, which would save a bunch of hassle right now...

ui/com/user-info.jsx
pfraze: Oh, no you can declare a constant function inside the constructor. Let me illustrate the pattern via es5:

```js
function foo() {
  function bar() {
    // stuff
  }
  bar()
}
```

So that's the basic pattern, a function within a function. And then, even in es5, you could do it as a variable assignment, like so:

```js
function foo() {
  var bar = function () {
    // stuff
  }
  bar()
}
```

So now, with es6, I use `const` instead of `var` whenever possible, to show when I do/dont expect a variable to mutate:

```js
function foo() {
  const bar = function () {
    // stuff
  }
  bar()
}
```

Then I use the fat-arrow syntax, partially for readability, but also for the `this` context:

```js
function foo() {
  const bar = () => {
    // stuff
  }
  bar()
}
```

The fat-arrow declaration changes the closure rules for javascript, so that it takes the `this` of the closure it was declared in. Without it, we'd have to do:

```js
function foo() {
  const bar = (function() {
    // stuff
  }).bind(this)
  bar()
}
```

or something similar. And then, lastly, we have the es6 class stuff, which is why we have a `this` to care about at all.

So here's what I was suggesting:

```js
class AutoRefreshingComponent extends React.Component {
  constructor(props) {
    super(props)
    const handleCB = (err, state) => {
      if (err) throw err
      this.setState(state)
    }
    this.computeState(null, handleCB)
    this.refreshState = props => { this.computeState(props, handleCB) }
  }
...
```

Probably more explanation than you needed, but I figured I'd err on the side of clarity :)

pfraze:
@mixmix the `app` object is put on the global `window` so that you can access it from the commandline. If you type `app.ssb`, you'll get the RPC interface to the main process. So, the devtools can do a bit of exploration for you.

You'll also notice in devtools, when the app starts, the manifest for `app.ssb` is emitted, so you can see all the functions that are available to you. You should be able to call `app.ssb.latestSequence()`

ui/com/user-info.jsx
mixmix: Ah yes, I knew you could declare variables inside functions, just feels strange and slightly counter my intuition for that to be a function. I think in Ruby it would be outside. Thanks for pointing that out. 

Thanks for the clear write up about `=>` and `bind` as well, I didn't know that detail explicitly

mixmix:
yeah I've seen the manifest and looked at where that comes from. I like following rabbit holes but feel like I lost the trail at some point. I think I might have stopped reading when it would have required jumping over to scuttlbot

mixmix:
selection_219

thanks for the review and advice @pfraze 

this is ready for another look

pfraze:
:+1: this is a really nice-to-have, and code looks great. Good work!

mixmix:
thanks!

PR #197, opened 2015-12-16
Add gpl-3 license

pfraze:
Based on the suggestions we've collected, I propose we merge this GPL3 license. If we find this is overly restrictive, we can alter the copyright in the future.

PR #227, opened 2015-12-30
Add custom About box, with license

clehner:
This adds an About window, to replace the default Electron one.
about
The dialog proactively assumes we make it to 2016. The icon is a resized copy of `assets/icon1024.png`, including a 2x version, according to the [NativeImage docs](http://electron.atom.io/docs/v0.30.0/api/native-image/).

The License button triggers another window. Currently the text assumes merging of #197.
license


pfraze:
:+1: awesome, love how you're pulling from package.json. Another great PR

clehner:
thanks. it could also be made to pull from the author & contributors fields of package.json - maybe in a separate "authors" list. Also the license title "GPL-3.0" will actually say "?" since #197 isn't yet merged. I guess I made that unclear with the screenshot

pfraze:
Cool, good to know. Authors would be cool to have too

PR #228, opened 2015-12-30
The New Years PR

pfraze:
This one got away from me a bit, since it started just from tinkering with a bunch of different things while I was on vacation. So we'll just call this the new years PR.

**Changes.**

---

**1** Moved much of the nav out of the top bar, replacing it instead with a search interface. Along with moving the nav, I renamed Feed->Public, and Inbox->Private. These changes are intended to reduce the visual complexity of the persistent UI elements.

screen shot 2015-12-30 at 12 52 54 pm

---

**2** There are now three viewing modes controlled at the top right: Inline (feed-like), Large (a big inboxish view), and Compact (a small inboxish view). Choosing this will persist across navigations.

Inline is above.

Large:
screen shot 2015-12-30 at 12 56 38 pm

Compact:
screen shot 2015-12-30 at 12 56 48 pm

---

**3** The search bar has a nice autocomplete "palette", which still needs some work but the basic ideas are there:

screen shot 2015-12-30 at 12 58 39 pm

---

**4** The composer still needs some aesthetic work, but it's more functional. I decided it couldnt stay inline at the top, because that style only worked with inline rendering. So, here's how it looks for private messages:

screen shot 2015-12-30 at 1 00 35 pm

Here's public messages, which now includes a channel control:

screen shot 2015-12-30 at 1 00 43 pm

And here's the reply control:

screen shot 2015-12-30 at 1 00 54 pm

That's it!


pfraze:
Fixed up that composer rendering:

screen shot 2015-12-30 at 1 18 30 pm

ui/layout.jsx
```diff
@@ -121,21 +98,13 @@ export default class Layout extends React.Component {
...
       <div className="toolbar titlebar flex">
-        <div className="flex-fill">
+        <div>
           <a className="ctrl back" onClick={this.onClickBack}><i className="fa fa-angle-left" /></a>
-          <div className="nav">
-            <NavLink to="/" selected={location === '/' || location.indexOf('/newsfeed/') === 0} icon="newspaper-o" label="Feed" />
-            <NavLink to="/inbox" icon="inbox" label="Inbox" count={this.state.indexCounts.inboxUnread} />
-            <NavLink to="/profile" icon="users" label="Contacts" />
-            <NavLink to="/sync" icon={isWifiMode?'wifi':'globe'} label='Network' />
-            <Issues />
```
clehner: Would it be useful to keep the Issues tab?

ui/com/search-palette.jsx
```diff
@@ -0,0 +1,115 @@
+'use babel'
+import React from 'react'
+import cls from 'classNames'
```
clehner: To support case-sensitive file systems, this should be 'classnames'

pfraze: good catch

ui/layout.jsx
pfraze: ah yes it would, thanks

PR #230, opened 2016-01-01
Thin-client Tracking PR

pfraze:
This PR puts patchwork back into the browser and ditches the electron dependency. Created in response to the the `%7WVt+O4xVYnOP10tADLO6RpTuBVRc1Do3eRbVjAtrCQ=.sha256` thread.

Things todo, if we choose to proceed with this PR:

 - [x] Restore profile image functionality
 - [x] Restore post attachments
 - [x] Replace the webview functionality
 - [x] Parse HTTP URIs into mentions, when creating a message
 - [x] Parse HTTP URIs into local links, when rendering a message (PR pending: https://github.com/ssbc/patchwork/pull/231). This will only be necessary for when the publishing-user's app doesnt do it when creating the message.
 - [x] Config option to expose Patchwork to the network
 - [x] Config option to password-protect Patchwork
 - [x] Config option to use TLS (PR pending: https://github.com/ssbc/pull-ws-server/pull/11)
 - [x] Electron-mode: Window menu
 - [x] Electron-mode: Context menu
 - [x] Electron-mode: Back button in nav menu
 - [x] Electron-mode: Multiple windows


pfraze:
This is ready for discussion and testing. It adds some new config items, which we'll need to put in the readme. They are:

 - `allowRemoteAccess`: if truthy, connections from IPs other than loopback will be allowed
 - `singleUserPassword`: if a string, connections will be required to provide the matching string through HTTP basic auth
 - `tls.key` and `tls.cert`: paths to TLS private key and cert, respectively. If provided, the application will serve over HTTPs.

Because of `rc` you can control these via the config or CLI. Here's my command to test all of them at once:

```bash
node index.js --tls.key ~/tmp/sslcerts/privkey.pem --tls.cert ~/tmp/sslcerts/newcert.pem --allowRemoteAccess --singleUserPassword bananas
```

pfraze:
Still need to clean up some parts of the repo and building (clear out all that electron corpse)

pfraze:
Ok done. @dominictarr this is ready for you to review, and for us to give some thought about. I ran it successfully on grimwire using a self-signed ssl cert

cryptix:
Connecting to `http://localhost:7777` on osx using firefox, I'm getting a lot of `Content Security Policy: ...` errors and the page just stays `Loading...` trying remote host now.

pfraze:
hmm yeah, I see why the CSP issue is happening. 2 secs

pfraze:
@cryptix I just pushed a fix (0ddbd2fb19e6d942989746cf524f424840770b87) see if that solves it.

The problem is that I was assuming `host` would only be explicitly configured by the user when they set up patchwork for remote access. In reality, if nothing is set for `host`, then it defaults to the public IP as reported by the OS. So, the CSP was coming back with your IP for the websocket's hostname, rather than `localhost`, which is what you needed.

cryptix:
@pfraze yes, that fixed it! some minor css hickups but that's for another issue, i think?

csshickup

pfraze:
Suggest we merge, any disagreements?

clehner:
:tada:

pfraze:
Published `2.1.0`

PR #237, opened 2016-01-06
Show recipient pics in PM threads

pfraze:
Updates the thread UI to show recipients in threads, via their profile pics.
screen shot 2016-01-06 at 2 00 26 pm

I tried this with hexagons, but the spacing is just really rough. Anybody have ideas how I could do it?

screen shot 2016-01-06 at 1 49 00 pm


pfraze:
Going to go ahead and merge, we can improve on this more if we want

PR #241, opened 2016-01-09
New setup flow

pfraze:
This PR updates the setup flow with something which is intended to be prettier, simpler, and more focused. It takes out the nearby-following and pub invites; since both of those are optional tasks, I'm going to solve them with prompts in the UI once they're through the setup.

Screen caps of the flow:

screen shot 2016-01-08 at 10 40 02 pm
screen shot 2016-01-08 at 10 40 09 pm
screen shot 2016-01-08 at 10 40 18 pm
screen shot 2016-01-08 at 10 40 46 pm
screen shot 2016-01-08 at 10 40 56 pm


jrf0110:
Ramblings on UI design just based on the screenshots

1. Welcome to _Patchwork_ screen is hot
2. Don't re-use that rainbow effect on the other screens. It feels like a crutch
3. Those icons underneath the progress indicator are pure UI fluff
4. I think a solid secondary color should be used for that progress indicator (I think something around `#AD48AA`)
5. The first screen is centered, but the rest aren't. I think maybe try centering the rest? The UI is so focused, and that's good, but it also means you can get away with putting the content "front and center"
6. Maybe a screen at the end that's like, "One moment while we load your data" or something so they don't see any data flicker
7. Love the "Choose your View" step
8. Maybe explain a little bit how usernames work on PW (how they're just aliases or some shit)

pfraze:
Followed your advice on 2, 3, and 5. Might do 4 at some point (wasnt feeling the purple though).

jrf0110:
Yeah, that purp was just a guess

pfraze:
screen shot 2016-01-12 at 2 46 32 pm
screen shot 2016-01-12 at 2 46 38 pm
screen shot 2016-01-12 at 2 47 06 pm
screen shot 2016-01-12 at 2 47 12 pm

jrf0110:
I'd definitely keep the text on the progress indicator; Just ditch the icon

pfraze:
I dunno, I felt like no text or icon made it feel really clean

bcomnes:
Looks great!

PR #242, opened 2016-01-09
refactor out markdown parser

dominictarr:
I refactored out the markdown parser, and gave it a few tests (that just check the current output) and docs.
https://github.com/ssbc/ssb-markdown

This makes it easy for other parts of the ssb ecosystem to render markdown posts consistently with patchwork. In particular, this would be of use to porthole, and other things people may or may not be building @NHQ

(Also, I needed this for a thing)


pfraze:
:+1: cool

NHQ:
neat-o

smartypants should be true tho so -- becomes an em-dash, cuz I don't know how else to make one.

NHQ:
BTW my PR for ssb-ref has checks going the other way, parsing plain text for mentin links.

Issue #114, opened 2015-11-23
Desktop notifications


pfraze:
Show desktop notifications for new posts that you care about. I suggest, for new direct mail, or replies to bookmarked threads.

PR #243, opened 2016-01-13
Desktop notifications

clehner:
This patch integrates desktop notifications. A notification is shown for each new thing that currently appears in the notifications sidebar. ~~To help with doing this, we also add a live-only option to the patchwork API methods that create streams, so that only new/live items get processed for showing notifications.~~

Other ideas:
- Add a UI settings option to enable/disable desktop notifications
- Notify in response to new direct mail, or replies to bookmarked threads, as proposed in #114

### DIFF
--- a/api/index.js
+++ b/api/index.js
@@ -524,6 +524,10 @@ exports.init = function (sbot, opts) {
       var readPush = pushable()
       var read = pull(readPush, paramap(fetch))
 
+      // allow live-only stream
+      if (opts && opts.live === 'only')
+        readPush.end()
+      else
       // await sync, then emit the reads
       awaitSync(function () {
         var added = 0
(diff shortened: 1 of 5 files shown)


pfraze:
Ok, sorry I'm taking so long to do review! Testing this now

clehner:
Squashing and rebasing

clehner:
Current issues with this feature:
- Notifications on Firefox sometimes don't work. It has something to do with the icon URL and [a firefox bug](https://bugzilla.redhat.com/show_bug.cgi?id=1250617)
- Occasionally, the state of the index gets reset or something, and notifications stream in from the beginning of one's feed to the present. This has happened twice to me so far, when I was switching between branches, and maybe rebuilt a dependency. So the live:only feed stream option needs to be fixed. Also maybe it would be a better API to call it {live:true, old:false}, so that the existing option remains a boolean.
- If Patchwork is open in more than one browser window, duplicate notifications are produced. This could be fixed by doing some synchronization between windows, or by moving desktop notifications out of the web code and into some native thing (growl, libnotify, etc).

PR #247, opened 2016-01-17
Newsfeed and inbox improvements

pfraze:
2 updates:
 - filter inbox and newsfeed to posts by followed users only
 - fix: newsfeed and inbox no longer accidentally skip chunks of the index

### DIFF
--- a/api/index.js
+++ b/api/index.js
@@ -550,7 +550,7 @@ exports.init = function (sbot, opts) {
           if (limit && added >= limit)
             break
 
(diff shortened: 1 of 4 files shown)


dominictarr:
this is quite a drastic change - is there a way to make this backwards compatible? or otherwise to make a gentle transition?

pfraze:
We were talking about it in PW. Yeah I think it'll need some more refinement, maybe a UI element to control, or maybe another view that doesnt do this filtering.

PR #248, opened 2016-01-17
keyboard friendlyness

krl:
So, i use a keyboard-centric browser, and patchwork did not highlight the links i wanted to press, so i made this. It looks a tad ugly, but using href, href='', or href="#" had behavioral side-effects.

It's not complete, dropdown and some onClicks on non-links still don't work. I'm not sure if having onClicks on divs is good practice.

This is also an accesability issue.

Discussion:

http://stackoverflow.com/questions/5637969/is-an-empty-href-valid


krl:
If you know a better way of solving this, please let me know.

I'm doing this against the hosted branch, since this is not really an issue with electron, that does not have keyboard support anyway. (i think?)

pfraze:
:+1:

PR #251, opened 2016-01-19
Tracking: use Naming plugin

dominictarr:
in progress PR to remove memory names table and use name plugin instead.
This is currently really slow because every view makes many requests to name plugin and currently that does the full db stuff each time, which is also slow because the way names work and the way indexes work are not optimized (but I have ideas for this)

naturally we should not merge this until any performance problems are fixed.


pfraze:
RE the multiple accesses causing slowness, I've had some ideas about that also. It's currently handled by the `app` state object, which is a pretty straight-forward solution but not super elegant. A few things about it worth noting:
 - It emulates the lifecycle of state that the web does, where each view-change triggers a "fetchLatestState"
 - It has an events pipeline that's been very ad-hoc

The app obj acts as a caching layer between the main process and the different components; it's how I've avoided the multiple db accesses problem. But there are more elegant ways to handle it. One is, we could wrap the RPC interface (`app.ssb`) with a caching layer, and add memoization. Another is, we could use something like immutablejs and create objects that sync with the backend, and emit a much less ad-hoc change events. So basically, idea A stays imperative, and idea B goes with data models. Might make sense to use both.

dominictarr:
There are lots of ways to approach this, but not general purpose way that is good for all cases.
With names, I think it would be okay to have some degree of memoization, I'm thinking things should be memoized for a while, but concurrent requests should make only one request behind the scenes.

I think that would make plenty of sense for this case, but maybe not for things that change more frequently.

pfraze:
Right, that's why I figure it'll probably be both option a and b. At any rate, I'd suggest you stick with the patterns in use now.

dominictarr:
Okay I updated ssb-names to use some caching - it just caches 100 names for 10 seconds max. But this is enough to make performance acceptable. This is enough to make reading names work, but now I need to update suggest-box to allow name suggestions to be async.

dominictarr:
turns out that suggest-box already supported async lookups, but it wasn't documented so I didn't realize that until I started try and add that feature, so it ended up getting refactored.

the reverse lookup of a name is slow, but it doesn't need to be if we changed how indexes work a little bit.

PR #252, opened 2016-01-19
Hosted webkit

clehner:
Fix PR adds gets the hosted branch working on webkit browsers (e.g. surf, Midori).
- Remove usage of ES6 features (Array.from and Symbol) from non-babelized code
- Handle less-featureful WebKit Error and Event objects
- Autoprefix CSS.

Unfortunately, adding autoprefixer makes building the CSS take 70% longer on my computer. However, it's still fairly quick (1.3s), and when the app is eventually packaged this shouldn't be a significant problem for the user.


ui/com/msg-view/summary.jsx
```diff
@@ -11,7 +11,8 @@ import u from '../../lib/util'
 export default class Summary extends React.Component {
   onClick(e) {
     // make sure the user didnt click on a link
-    for (var i=0; i < e.nativeEvent.path.length; i++) {
+    var pathLen = e.nativeEvent.path ? e.nativeEvent.path.length : 0
```
pfraze: I dont think we can just default to 0; that would cause this code to fail. This piece would have to be rewritten to recurse up the nodes by following `.parentNode`.

pfraze:
Looks good except for my one note. That sucks about the longer less compile time :frowning:. Might be worth doing a "build for release" that's different from just "build for dev testing"

ui/com/msg-view/summary.jsx
clehner: Attempting to fix now. I think it works.

clehner:
Build for release could do some CSS minifcation maybe.

pfraze:
> Build for release could do some CSS minifcation maybe.

yeah

pfraze:
:+1: Changes look good, nice work

clehner:
Thanks

PR #253, opened 2016-01-20
Fixes

clehner:
Here are some fixes for things.
- Work around event.path on non-Chromium
- Fix error about peers when disconnected from sbot.


pfraze:
:+1: solid

Issue #130, opened 2015-11-27
Make history syncs efficient


pfraze:
I've gotten a couple reports that, when using the first invite, Patchwork becomes unresponsive. When you consider how Patchwork handles incoming messages, this is not surprising.

The following happens when you open the newsfeed view:
 - The UI process opens a live stream to the main process' newsfeed index. 
 - As each log-entry comes in, the main process updates its indexes and views. If a log-entry updates the newsfeed index, the main process emits the new/updated message-thread on the live stream
 - The UI process receives the new/updated message-thread, and puts it in a queue. The user is notified of a new update, and can click to render the changes.

Thankfully the feed doesnt just auto-render anymore - each update goes into the queue, for batch rendering when the user clicks. But, there's still some transmission and processing overhead for each update. There's also memory cost, as you end up loading the entire history into the queue. Then, god help you when you press the button to render.

So here's the problem: the behavior here is just fine when you're rendering realtime updates. But, it's quite bad when you're syncing history. And more troubling is, you can end up syncing history at any time; this isn't just something that happens when you use an invite. Any time you follow somebody new, you'll sync history.

So, we need to come up with a way to keep newly-received history from being treated like realtime updates.

pfraze:
Labelling this as a bug, because it's a really bad experience when it occurs.

PR #262, opened 2016-01-22
suppress index-additions for rows with a TS more than 48 hours from present time

pfraze:
Solves https://github.com/ssbc/patchwork/issues/130. I'm not sure this is the right solution, but it works. Only posts with an asserted timestamp within the last 2 days will emit in the live-stream.


pfraze:
Still not sure this is the right final solution, but it's a pressing bug to solve so I'm going to go ahead and merge

PR #263, opened 2016-01-22
Composer refactor

du5t:
much cleaner than PR 4. bonus: fixed secret regression! also removed Editor for now

### DIFF
--- a/ui/com/composer/index.jsx
+++ b/ui/com/composer/index.jsx
@@ -42,7 +42,7 @@ class ComposerTextareaFixed extends React.Component {
 }
 const ComposerTextareaVerticalFilled = verticalFilled(ComposerTextareaFixed)
 
-export default class Composer extends React.Component {
+class CompositionUnit extends React.Component {
   constructor(props) {
     super(props)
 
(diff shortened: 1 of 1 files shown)


ui/com/composer/index.jsx
```diff
@@ -214,10 +379,15 @@ export default class Composer extends React.Component {
     mentionslib.extract(text, (err, mentions) => {
```
du5t: by the way, this is repeated a lot in Editor also (prepping the text to send), but I'm not quite sure how would be best to combine yet, since sending goes on in the callback.

pfraze:
:+1: Had to patch up a few things, but otherwise good.

In the future, it's fine to refactor the code, but don't reformat large sections of the code (all the multiline stuff). It made the PR harder to review, and it introduced some rendering errors because of the way React handles spaces and newlines.

du5t:
Sorry about that; thanks for fixing. I have a ton of trouble reading JSX, but I guess I should have saved the style conversation for after getting a feature through.

pfraze:
No worries. I have mixed feelings about multiline jsx like that... I wont reject a PR from you if you do it on new code, though.

PR #265, opened 2016-01-22
fix: a bunch of style updates to solve flexboxes going too wide

pfraze:
Seems that newer browsers changed their interpretation of flex-boxes somehow, and that caused a bunch of width overflows. This solves it.

### DIFF
--- a/ui/com/msg-content.jsx
+++ b/ui/com/msg-content.jsx
@@ -30,7 +30,7 @@ export class Inline extends React.Component {
     let c = this.props.msg.value.content
 
     if (this.props.forceRaw)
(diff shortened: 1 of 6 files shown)


dominictarr:
flexbox sounded great when I first heard about it but turned out to be much more difficult than I thought, especially when I tried to create a flexbox that had a overflow scroll but I wanted to fit inside the screen. I ended up going back to setting the height directly with javascript.

pfraze:
I'm still happy with it overall, but yeah, there are some serious WTFs with it. The properties in general are not intuitive. I have no idea what the designers had in their heads.

dominictarr:
the usual story

PR #266, opened 2016-01-22
styles: wider messages

pfraze:
Messages seem a little narrow to me, leaving a lot of blank space where we could have content. This PR goes from this:

screen shot 2016-01-22 at 12 49 59 pm

To this:

screen shot 2016-01-22 at 12 49 10 pm

And also closes the gaps between messages inside a thread:

screen shot 2016-01-22 at 12 49 13 pm

### DIFF
--- a/ui/less/com/composer.less
+++ b/ui/less/com/composer.less
@@ -76,7 +76,6 @@
       resize: none;
       border: 0;
       min-height: 140px;
-      max-width: 586px; // mimics the rendered cards
       padding: 10px;
       margin: 0;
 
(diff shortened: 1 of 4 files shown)


pfraze:
Im toggling between these to see which I think is easier to read. Not sure yet.

dominictarr:
A column of text should be about the same as in a paperback book. about 10 - 15 words across. If the lines are too long then it's harder to scan back to the start of the next line. I think the wider one is acceptable, but I don't think there is any problem with the narrower one either.

pfraze:
ok I think it's easier to read when it's a little thinner, so I'm going to close this

Issue #173, opened 2015-12-10
Improve all load performance 


pfraze:
Feed load, feed scroll, and thread load are all too slow. This can be improved in rendering, and with better processing and caching. Make it so!

pfraze:
Some opportunities I've found:
 - Setting env var `NODE_ENV="production"` disables some React debugging that shaves off some ms
 - [shouldComponentUpdate](http://facebook.github.io/react/docs/advanced-performance.html) is essential to getting small updates and scrolling to be jank-free.

Rendering 30 messages on the newsfeed purely from memory takes ~80ms. From SSB, takes ~250-300. I need to chart performance of the pure API loads, and then of the IPC transfers. There should be some improvements to make in there.

The last resort solution is to cache messages inside the UI thread.

pfraze:
https://facebook.github.io/react/docs/perf.html react perf tools

dominictarr:
what about streaming the load, it's nice to have 30 mesages because I will scroll down in a second, but I can't read that fast. would showing the first message asap be faster?

pfraze:
I've been planning to try that again. In previous tests, the render performance for streaming in posts was really poor. But, since then, I discovered the `shouldComponentUpdate` optimization could solve the slowdown.

dominictarr:
hmm, what is `shouldComponetUpdate`? that sounds like a leaking abstraction.
maybe react isn't ideal for streamy appends, but is really for easy to maintain arbitary updates.
If we go with something to specially manage the infinite scrolly bits (feeds) how would that integrate with react? can we add react components into a thing which is not react? can we just add one more react element?

pfraze:
It's an optimization to deal with a component that has a lot of vdom to diff (which the msg-list does). I think it's better to use it, than to try to escape out of React for this piece.

pfraze:
Both of those improvements were implemented. Performance is an ongoing task, but I'm going to close this issue

PR #268, opened 2016-01-23
Feed frontend render speed improvements

pfraze:
Per https://github.com/ssbc/patchwork/issues/173, this PR improves on feed rendering on the frontend by utilizing `shouldComponentUpdate` and by stripping out all of React's development-mode stuff in the production build. Both of these merit some comments.

To make shouldComponentUpdate work, I added a "change counter" to the message objects. Any time the msg-list component makes an update to a message that will need rendering, it increments that number. The msg-view component then observes that `changeCounter` has been altered, and tells React to that `shouldComponentUpdate == true`. It would be more elegant if we used immutable-js or something like it, but this solution works, and the perf improvement is big enough to warrant this.

To strip out React's development warnings, I added envify to the build process, which replaces all `process.env.NODE_ENV` references with the actual value. I then added uglify, which removes dead code (which, in this case, is everything that's not supposed to run when node_env isnt "production"). Because of the tests and try/catches that are removed, I actually got a 20-50ms boost in page loads from that.

It's all faster, but it's not amazing yet. For me, the newsfeed renders between 270-330ms on average, and there's a slight jank when the feed loads more messages. I'd like it to be down to 150ms, at most, and have no jank at all.


pfraze:
The scroll jank is a single jerk after messages are loaded. Since more messages are loaded before you hit the bottom of the screen, I can see that the jerk isnt caused by actually hitting the bottom - it's from the moment of rendering. Not sure why that is, yet.

Issue #273, opened 2016-01-27
Problem with incremental build


clehner:
When I run build:js or build-release:js, I often get this error in the browser on load:
> We're sorry! Patchwork experienced an error while loading.
>
    TypeError: undefined is not an object (evaluating 'require("../package.json").version')

The code `require("../package.json").version` is from the `elliptic` module, and it seems that browserify or browserify-incremental have trouble resolving it sometimes.

I can get around this is by deleting `browserify-cache.json` and having a slow build. I have tried with browserify 13.0.0 and 12.0.0.

pfraze:
ah... that is weird and unfortunate, especially because 1) it doesnt happen to me, so that's weird to reproduce, and 2) slower build

hm. we can either ditch browserify-inc, or add a "--clean" flag to the build steps to optionally clear browserify-cache.json. I figure the latter is better since it does seem to work sometimes...

clehner:
If it's just me, I wouldn't bother adding a special case. I wanted to post this here in case other people experience the same issue. I will see if I can fix the underlying thing in browserify. Also I am trying out using watchify for incremental builds, which I find to be pretty fast.

pfraze:
ok, let me know what you find (including w/watchify)

Issue #255, opened 2016-01-21
Handle RPC connection death


pfraze:
If the server is closed, the UI should put up an error screen explaining that the server died. Should include instructions to restart the server, then refresh the page.

pfraze:
I should clarify, this issue is in reference to the local server, not pubs.

pfraze:
For somebody looking to work on this issue, a simple and useful start would be to create a nice modal which I can show when the connection is lost. If you refer to `ui/layout.jsx`, you'll be able to see how the setup modal is inserted into the page and controlled. You can use `ui/com/modals/single` in this case. Good luck!

pfraze:
This got done

Issue #272, opened 2016-01-27
Build fails on windows due to environment variables


pfrazee:
`NODE_ENV=development node ...` doesnt work on windows, causing the builds to fail

pfraze:
There are also issues with nix-specific commands, like mkdir -p

ahdinosaur:
hey. i'm closing this since stale due to [`v3` release](https://github.com/ssbc/patchwork/releases).

in other news we magically [just got `v3` to build on windows](http://viewer.scuttlebot.io/%25%2BSBr3lvHbR2yNKASU7FfLX7fTN06yk4ON1qUipaG9fk%3D.sha256)! :tada:

PR #279, opened 2016-01-28
Add watchified build scripts

clehner:
This adds auto-rebuild scripts for the JS, CSS, server (auto-restart), and one to run all three.

This also moves the build script from the separate script file into the package.json, using browserify's JSON syntax for the transforms. This allows the config to be shared by browserifyinc and watchify, and results in less code. Some testing should be done on this to make sure that this does not cause a regression of the issue of babel finding presets when symlinked node_modules are used. (It hasn't yet for me).

nodemon is used for restarting the server (which currently requires the client to refresh to reconnect (#255)). I could not find any LESS-specific tool for rebuilding the stylesheets incrementally that worked with having a single entry-point less file importing other less files, but it's quick enough to build them in one shot and nodemon handles running arbitray commands and watching arbitray files, so we use it here for auto-rebuilding the CSS as well.

Ref https://github.com/ssbc/patchwork/issues/273#issuecomment-176297535 and
[%DkiS2KWRQdkMb9U64njPVeFZ4InKKZfIEtBCNMtmhYM=.sha256](http://localhost:7777/#/msg/%25DkiS2KWRQdkMb9U64njPVeFZ4InKKZfIEtBCNMtmhYM%3D.sha256)


pfraze:
Nice. Funny you moved the build out of JS, because https://github.com/ssbc/patchwork/issues/272 makes me think we may need to move them all into JS for portability. We'll see, but this should be :+1: for now

mixmix:
nice!

clehner:
One problem I have noticed with this is that sometimes watchify has recognizing jsx files. I find it happens sometimes when switching branches, I will start seeing errors like this:
```
/usr/local/src/patchwork/ui/routes.jsx:2
import React from 'react'
^
ParseError: 'import' and 'export' may appear only with 'sourceType: module'
```
When this happens I have to restart watchify. This takes a while, so it would be good to fix this problem.

I think there may be some config option to fix this but I haven't found it yet.

clehner:
Ok, I found that the problem I mentioned was happening because when I was switching branches, I was switching to a branch that didn't have the browserify transforms in package.json, and so browserify or babelify was rereading package.json and not getting the correct options. If the changes in this branch are merged into the other branches then this problem does not occur.

So unless anyone else has issues with this PR, it should be good to merge.

pfraze:
Ok :+1: I havent tested yet but easy to fix if something's wrong, and Im sure there wont be anyway

PR #285, opened 2016-02-01
Electron fixes

clehner:
- Show Window menu
- Fix About window missing image
- Add link to website in About window

~~For some reason, when I build a package using the patchwork-electron release scripts, the image in the About window is not recognized and was giving me an error, unless I ran the executable in the directory of patchwork. However, by wrapping the image path with nativeImage(), the error is suppresed and a generic icon is used instead, which fixes that problem partially.~~ Edit: Fixed in 3758bc0; nativeImage() not needed.


pfraze:
:+1:

PR #291, opened 2016-02-06
Feature/copy-id

mixmix:
copying id wasn't working. 
Now works (minimally) in electron and firefox.

@pfraze would love some input on how i'm managing passing the parent to the child menu components so that I can modify state ... I tried refactoring what was there a little, but not sure what good style is yet.

Known problems - toggling raw view state doesn't seem to work reliably. Works more reliably than it used to, but not sure how to debug problem...

### DIFF
--- a/ui/com/msg-view/card.jsx
+++ b/ui/com/msg-view/card.jsx
@@ -98,10 +98,6 @@ export default class Card extends React.Component {
       this.setState({ isViewingRaw: !this.state.isViewingRaw })
   }
 
-  copyLink() {
-    prompt('Here is the Message ID. Press cmd/ctrl+c to copy it.', this.props.msg.key)
-  }
-
   componentDidMount() {
     // load subject msg, if needed
     let msg = this.props.msg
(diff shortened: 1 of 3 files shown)


pfraze:
I'll make some quick changes

mixmix:
Thanks. Comments would be appreciated.

mixmix:
If you have time, otherwise I'll check out commits

pfraze:
Sure

pfraze:
First update, the new per-item behavior for clicks was being used exclusively, and not falling back to the props.onSelect if unspecified.

pfraze:
The data view failing to render had to do with an existing bug in `shouldComponentUpdate`.

pfraze:
@mixmix why did you introduce the whole `faClass` thing?

pfraze:
Oh, I think I see why. Ok, I prefer a different approach to that

mixmix:
It was an attempt to dry up the code some, there was formerly a lot of repetition of structure. I extracted the unique parts. 
Interested to see your pattern

pfraze:
In this case, I prefer to leave it as it was. Leaving label as a more general attribute bakes in fewer patterns and stays flexible.

pfraze:
I didnt want the ClipboardBtn inside the dropdown, because that was inter-tangling a specific use-case with the dropdown, which should be more general. So, I added the `Com` attribute for providing a custom component on a dropdown option.

Unfortunately the ClipboardBtn also didnt call `stopPropagation` on the click event, so I had to rearrange the dom elements to keep clicking on the Copy ID btn from "reopening" the dropdown after it closed.

pfraze:
Ok, I think this is ready to go now

clehner:
@pfraze can I merge this into smarter-inbox-pt3?

pfraze:
@clehner sure, though I figured it would be easier just to let it merge in when smarter-inbox-pt3 goes into master

clehner:
Well, it's there now. Also the new unread state looks good.

pfraze:
cool

ui/com/msg-view/card.jsx
```diff
@@ -214,13 +227,24 @@ export default class Card extends React.Component {
...
-      { value: 'copy-link',  label: <span><i className="fa fa-external-link" /> Copy ID</span> },
-      { value: 'toggle-raw', label: <span><i className={isViewingRaw?'fa fa-envelope-o':'fa fa-gears'} /> View {isViewingRaw?'Msg':'Data'}</span> },
-      (isDownvoted) ?
-        { value: 'unflag',   label: <span><i className="fa fa-times" /> Unflag</span> } :
-        { value: 'flag',     label: <span><i className="fa fa-flag" /> Flag</span> }
+      {
+        value: 'copy-link',
+        Com: props => <ClipboardBtn component='li' data-clipboard-text={msg.key} onSuccess={props.onClick}>
+          <i className="fa fa-external-link" /> Copy ID
+        </ClipboardBtn>
```
mixmix: @pfraze this is pretty interesting. When this `item.Com` is used later, it doesn't get called like `item.Com(this.props)` 
Is that something that is implicitly done when jsx is transpiled  ?

i.e.
```jsx
<item.Com key={i} onClick={onClick} />

// becomes something like: 
<item.Com(this.props) key={i} onClick={onClick} />
```

tried googling stuff like _dynamic jsx component function_ ... not that much luck. If you've got keywords that would improve my search those would be great.

pfraze: Those are called stateless component functions, they're written about in this blog post https://facebook.github.io/react/blog/2015/09/10/react-v0.14-rc1.html

PR #294, opened 2016-02-09
Fix error on closing composer

clehner:
If I click the Cancel button to close the composer within a few seconds after opening it, I get an issue:
```
TypeError: undefined is not an object (evaluating '_this2.refs.composer.style')
```
This patch clears the timer set on opening the composer so it doesn't get fired when the composer is closed.


pfraze:
:+1:

pfraze:
published `2.3.1`

PR #298, opened 2016-02-09
Make post date a link to the post

clehner:
This patch makes the date text on a post be a link to the post. This makes it easier to get a link to a post. It also makes it possible to navigate to a thread/post from the search results.


dominictarr:
:+1:

PR #302, opened 2016-02-16
Render raw msg json

mixmix:
Here's an alternative implementation for the raw msg data view

selection_263


pfraze:
Yeah I like this. I'm unsure about putting `id` on the object. In the code, messages always look like this:

```
{ "key": ..., "value": <the object above> }
```

That's because `key` is the content hash of `value`. But it's a total pain in the ass to do that - what you have in this PR with `id` on the object is much more convenient to look at and reference.

Hum. I'm going to go ahead with this anyway.

mixmix:
Sweet!
Let's close that other pr (if haven't already)

PR #306, opened 2016-02-18
Bringing back the feeds

pfraze:
screen shot 2016-02-18 at 3 12 33 pm

 - Makes the home page and channels views a feed
 - Orders the home & channels by original post time, not by reply times (you cant bump by replying)


dominictarr:
:+1:

PR #313, opened 2016-02-22
Refined inbox, updated nav, and reintroduce feed

pfraze:
Three major changes:

 - Inbox now clears out read items, putting them behind an "archived" button
 - Left and right navs are updated, and now have collapsable menus
 - The feed-view is back, and incorporates everything from #306 

The flow is setup so that you do work in your inbox (your priority queue), and then you can go into the activity stream and channels to discover new content, and potentially pull that content into you inbox (by watching the threads).


pfraze:
Note, also, the activity stream and channels are now sorted only by the first message's publishing. My reasoning for this is, if you want to follow something, you can watch the thread. Otherwise, the new threads are generally new content, and so lets have new threads go at the top.

dominictarr:
hmm, that is _definately_ a breaking change, but on the other hand, it's gonna simplify the view/query.
I predict that will hamper engagement, because most people won't be reminded about old theads that are updated, unless they actively watch them, which since it's active is probably less.

pfraze:
@dominictarr yeah and it may be something we need both of. Might need a new-sort/active-sort toggle.

pfraze:
@dominictarr humm. Yeah Ive been trying it with this new order and Im not satisfied with that either! it never ends.

Maybe that ordering shouldnt change, and then there's an optional new view... Also, something else I've been thinking about is "watching a channel" which would cause all messages on that channel to auto-bookmark and go in your inbox.

pfraze:
Ok, I reverted the public/channel orders to as they were, so Im going to consider this a minor bump

pfraze:
published `2.5.0`

PR #319, opened 2016-03-05
add right-padding to markdown table cells

mixmix:
Before:
selection_286

---

After:
selection_285

... just a little bit less cluttered. In particular the table headers, which would no longer need that '/' hack to not look like a mess

### DIFF
--- a/ui/less/com/markdown.less
+++ b/ui/less/com/markdown.less
@@ -29,4 +29,8 @@
     cursor: default;
     text-decoration: none;
   }
+
+  td, th {
+    padding: 0 6px 0 0;
+  }
 }


dominictarr:
@mixmix your before and after pictures are the same. surely that isn't right?

clehner:
@dominictarr the difference is in the margin surrounding the table header cells. The change makes it so I wouldn't have had to put the slashes to separate the contents of the header cells.

mixmix:
It's minimal so it doesn't wreck tables with more columns. Trying to go for
more readability

pfraze:
:+1: looks good to me

PR #329, opened 2016-03-07
Quick bugfix & tweak

pfraze:
- Fix a bug where composer would give the wrong `branch` link
 - Always expand the root post when viewing a message *page*


pfraze:
published `2.6.3`

PR #332, opened 2016-03-13
Add home view, with merged inbox and activity feed

pfraze:
It was bugging me that I had to switch between the inbox and activity feed pages. So I created a new Home page, which puts the unread inbox items at the top of the feed:

screen shot 2016-03-12 at 7 17 27 pm

If there are no unread items in the inbox, then this area is empty.

Looking for feedback on this change.


tinkerb:
IMO, this is not a good idea. It blurs the line visually between what is public and private, which I predict will cause many users (especially new ones) to feel a sense of confusion and uneasiness.

If this is something that is definitely going to happen, I would recommend some kind of visual differentiation between public and private content, like different color cards (e.g. a new dark gray for private content while public content remains light gray).

pfraze:
Nah, nobody seems to like this. Closing.

PR #335, opened 2016-03-24
Set activity stream as the main page

pfraze:
@dominictarr pointed out that the inbox wasnt the right main page, and I agree. This sets the activity stream to the main page


pfraze:
I also snuck in a commit to make the leftnav be expanded by default (first load)

pfraze:
Anybody object?

PR #338, opened 2016-03-24
Merge digs and follows into one notices view

pfraze:
I didnt like having follows/unfollows in my inbox, so I just moved them into the digs view and called the merged view, "notices"


pfraze:
Anybody object to this?

PR #340, opened 2016-03-24
Only allow standard ascii in channel names

pfrazee:
Some people had been sneaking gibbering into the channel names. I figure it's better to just normalize them, for now.


pfraze:
@dominictarr any objections to this?

kyrias:
Being non-American/non-English, there are quite a few instances I can think of where limiting it to ASCII-only is sort of annoying.  But I do agree that it probably shouldn't be completely free-for-all.

api/processor.js
```diff
@@ -404,9 +407,11 @@ function nonEmptyStr (str) {
   }
 
 // allow A-z0-9._-, dont allow a trailing .
-var badNameCharsRegex = /[^A-z0-9\._-]/g
-function makeNameSafe (str) {
-  str = str.replace(badNameCharsRegex, '_')
+var badNameCharsRegex = /[^A-Za-z0-9\._-]/g
```
ralphtheninja: Any good way of refactoring this regexp into a function? Seems it's used at five places.

pfraze: yeah I've been meaning to move it into patchkit-util

ralphtheninja: > yeah I've been meaning to move it into patchkit-util

:+1:

PR #345, opened 2016-03-26
add simple drafts

pfraze:
PW will now retain your draft in the text editor, on navigation. It differentiates between composing a new thread, and replies.


dominictarr:
oh just saw this... if you are only saving text, won't it loose mention state? what if I rename someone in between edits? I think this would be fine if all the state was recorded in the text, (for example, via markdown links) `[@name](@<id>)`

pfraze:
Nope, the mentions are extracted at the time of publish, so this works as expected

PR #352, opened 2016-03-29
add box-shadow to over-sized content

mixmix:
The intention here is to suggest oversized content is [subducting](https://en.wikipedia.org/wiki/Subduction). 

selection_316

### DIFF
--- a/ui/less/com/msg-view/card.less
+++ b/ui/less/com/msg-view/card.less
@@ -41,8 +41,8 @@
     }
   }
   .body {
-    margin: 10px 20px 0;
-    overflow: hidden;
+    margin: 0;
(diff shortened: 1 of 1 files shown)


mixmix:
there's a known problem with this - when there's an image in the box and the image is white, what I've made looks shit: 

selection_318

Other images look fine:

selection_317

This explains what I think the problem is. I tried implementing, but had trouble : 
http://stackoverflow.com/questions/21414925/why-doesnt-inset-box-shadow-work-over-images
Open to thoughts

mixmix:
moving this to : https://github.com/patchkit/patchkit-msg-view/pull/1

PR #356, opened 2016-03-31
Handle notfound msgs

pfraze:
There's been a (fortuitous) situation where, for some reason, Im not receiving a FoaF's feed in a timely fashion. (That's probably a bug!) But, the good news is, it's letting me see what happens when there are a lot of missing messages.

This PR adds some defensive guards against missing messages (sanity checks). I decided it's simplest to disallow replies on a thread that's missing it's root, because the typical logic is to read metadata from the root (channel, recipients, is encrypted?, etc)

I also updated the inbox index so that it doesnt add missing messages:
 - If it's a private message or mention, and the root isnt available, the message that *is* available will go in your inbox. That would be, the reply you just received, or the message that mentioned you.
 - If it's a message in a channel you watch, it wont get indexed at all.

### DIFF
--- a/api/processor.js
+++ b/api/processor.js
@@ -16,7 +16,17 @@ module.exports = function (sbot, db, state, emit) {
       // inbox index:
       // add msgs that mention or address the user
       var inboxRow
-      if (u.findLink(mentions, sbot.id) || u.findLink(recps, sbot.id)) {
+      if (u.findLink(mentions, sbot.id)) {
+        // if a reply, make sure we have the root already
+        if (root)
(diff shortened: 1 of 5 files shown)


pfraze:
Some things I suggest we discuss on next call:

 1. What's going on with this FoaF not syncing? (Probably just a protocol bug?)
 2. Is it time to get serious about individual message syncing?
 3. What's the right way for Patchwork to react to missing root messages? (As opposed to my bandaids)

cc @dominictarr @clehner @mixmix

PR #370, opened 2016-04-25
use ssb-blobs module instead of scuttlebot's built in blobs plugin

dominictarr:
this branch tests sbot with the new blobs plugin - now blobs are requested on demand - so for one, this won't download everything in the case of a new sync.


dominictarr:
unless patchwork attempts to render every message? @pfraze

pfraze:
Are you going to remove the old blobs code?

pfraze:
Only time blobs get requested in Patchwork is for rendering profile images, or embedded images, or when somebody clicks the link to a blob.

pfraze:
The multiple-hop want/have gossip solution looks promising.

dominictarr:
yeah, to really test, exit patchwork `cd ~/.ssb; mv blobs _blobs; mkdir blobs`
then restart patchwork. since patchwork already checks `blobs.has`

clehner:
i'm testing this now and it seems to be working. are there any concerns before i merge this?

dominictarr:
I was thinking the next step would be to have pubs that use both new and old strategies (and then automatically disable old after some point). but it should be fine to merge this into patchwork.

PR #371, opened 2016-04-27
nvmrc with less specific version that works

ahdinosaur:
i already have 5.3 installed, and this works, why force me to use 5.4? :)


pfraze:
👍

PR #372, opened 2016-05-02
set default background-color for user input elements

ansuz:
I'll start by saying I'm totally open to doing everything I did in a different way, but this PR has some minor improvements (for those running on a somewhat sketchy platform).

I read through some of the less files, and set these new values where I thought they made sense.

My problem was that gnome's dark theme gives firefox input and textareas a black background, which made user input fields practically unreadable in poor lighting.

I found two seemingly appropriate files for adding style rules for these elements, and set their values to use `@color-user-input`. This value is defined in a new file I created (colors.less).

My thought was that if we start to put colors into one file, it will be easier in the future for people to reskin their client by tweaking these values. I used 'white' for a background.

Thoughts?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ npm-debug.log
 ui/css/main.css
 ui/main.build.js
 browserify-cache.json
+.*.swp
--- a/ui/less/colors.less
+++ b/ui/less/colors.less
@@ -1,0 +1,1 @@
+@color-user-input: white;
--- a/ui/less/colors.less
+++ b/ui/less/colors.less
@@ -1,1 +1,0 @@
-@color-user-input: white;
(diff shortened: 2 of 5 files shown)


Kubuxu:
I am also affected by same problem where using dark GTK style makes input boxes dark inside.

pfraze:
I'm 👍  this change, but let's get rid of the colors.less. Variables either need to be used consistently or not at all.

ansuz:
Changes are now implemented directly in the files without colors.less.

I think it would be a nice idea to have colors in one place so people can tweak their installation a little more easily, but I'd much rather have the changes land :) Maybe later?

pfraze:
Yeah that's fine, it just needs to be consistent

pfraze:
thanks! 👍

ansuz:
:smile:

PR #377, opened 2016-05-03
Load scuttlebot plugins

clehner:
This allows for easier development and integration of scuttlebot plugins.

Depends on ssbc/scuttlebot#327


pfraze:
Looks 👍

PR #384, opened 2016-05-14
update ssb-config

dominictarr:
we need ssb-config to be at v2 to get long lived connections


dominictarr:
@clehner can you publish this? this is worth a patch version

clehner:
published 2.8.2

dominictarr:
awesome, thanks!

PR #389, opened 2016-05-31
display images with orientation in image's exif data

dominictarr:
this is one of those things that should problably just be the default but here we are

### DIFF
--- a/ui/less/main.less
+++ b/ui/less/main.less
@@ -1,6 +1,9 @@
 body {
   background: #eee;
 }
+img {
+	image-orientation: from-image;
+}
 
 @import "node_modules/patchkit-base-styles/all.less";
 @import "elevations.less"; // TODO remove


pfraze:
nice

mmckegg:
@PrasannaVenkadesh 

I'm assuming you use patchwork-classic from firefox? As far as I know, this doesn't work in chromium based browsers (including Electron).

In new Patchwork, it handles the exif tags on the upload instead. So if you add a blob with either Patchwork or (I think) Patchbay, it will autorotate correctly. It's only if you originally uploaded the file with patchfoo or patchwork-classic that it is incorrect.

This is a weird little problem! Wish chrome would just support auto exif rotation!

dominictarr:
maybe patchwork@3 is rotating the image but keeping the old exif?

PR #391, opened 2016-06-23
Enable Running on Windows

interfect:
SSB now starts up on Windows. It looks like the main limitation is
Chloridedown not being able to build, because it blithely calls "make"
as its build process.


dominictarr:
:100: great. it should still work without chloridedown, just using the javascript version (a bit slower but hey)

clehner:
:+1: and the build still works on Linux

dominictarr:
if this works we need a windows build on the releases page

PR #392, opened 2016-06-23
Fix local peer detection

clehner:
Decide if a peer is local based on whether it was found locally (vs. from a pub announcement) rather than based on whether its IP address is considered local. This fixes the issue of peers with cjdns IP addresses appearing as local as discussed in `%WUeEF57h1/YmMvNKsvv15sV/N9mKtOFoHJUbE0VKoKE=.sha256` and `%+kql+KJMI1m4KKM0ejZJSmQoA7JKEMz212TyAdIEhWI=.sha256`

### DIFF
--- a/ui/views/sync.jsx
+++ b/ui/views/sync.jsx
@@ -29,9 +29,7 @@ function peerSorter (a, b) {
 }
 
 function isLAN (peer) {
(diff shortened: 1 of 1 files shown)


pfraze:
Nice solution

PR #404, opened 2016-07-22
Unknown sources no longer required

balupton:
as electron binary now signed


clehner:
thanks. merging this and #403.

PR #411, opened 2016-08-26
update deps

dominictarr:
this updates scuttlebot, and ssb-keys to latest.

this is a patch change. We should also do a release, because 2.10 hasn't been working for some people


clehner:
Merged into 2.10.1

PR #419, opened 2016-09-29
fix display of "Digs on your posts" menu item

mmckegg:
The "Digs on your posts" menu item is currently showing as "NoticesMenuItem" due to a typo. This PR fixes it so that it uses the localisation as intended!


dominictarr:
awesome, merging

Issue #424, opened 2016-10-27
electron: Super slow start-up [caused by saturating master thread]


mmckegg:
I've been noticing start-up times over the last few months deteriorating massively. Sometimes it locks up for multiple minutes.

Initially I thought that this was caused by the render thread, but on closer inspection (well actually after I experience the same problem in [patchwork-next](https://github.com/mmckegg/patchwork-next)), I realised it must be something to do with self hosting the scuttlebot.

So it looks like what's happening is electron relies on having the "main" thead available for doing UI updates in all render threads. **If the master thread gets saturated, all of the render threads lock up.** Something to do with electron's synchronise IPC I suspect.

On initial startup, scuttlebot connects to many different pubs, and pretty much is tight looping for a good portion of that replication process (at least for me). I hacked around this in patchwork-next by changing initial gossip behavior to replicate with less clients at once on start up, rolling out slower.

But probably a better solution would be to move the self hosted scuttlebot out of the main thread and into another process (probably a hidden render process would do it fine - in Ferment I'm already doing this for torrents).

dominictarr:
yeah, I've been looking at this - what is happening is it by default, it connects to ALL the pubs on start up.
which hammers the database, but worse, it also uses up quite a bit of bandwidth!

The plan I have been working on is to persist the state of the peers, so that it's not necessary to reconnect to them all immediately (because the main reason to do that was to see which ones where actually alive)

it's not finished, but starting on this is `persistent-gossip` branch of scuttlebot

PR #425, opened 2016-10-28
fix slow start up problems caused by main process saturation

mmckegg:
**This PR splits out the scuttlebot server into a seperate render process when running inside electron. This improves the problems as outline in #424**

It also slows down the rate of initial connections to pubs. You connect to pubs that you follow immediately, but others are rolled out over a few minutes. This keeps the server responsive during initial load (otherwise endless scrolling won't work until all of the replication is done). 

It's all a bit hacky, but it does seem to do the trick!

@dominictarr, @ev: can someone add me as a collaborator on this project? Also I wouldn't mind the npm publish rights ☺️


mmckegg:
Wow it's so nice to be able to use Patchwork without constant glitches again! 🎆 🌈 

And it's not just startup. Everything is much smoother now. There used to be a lag whenever I pushed "Like", now it's instant!

I actually wonder if some of this was caused by upgrading to the latest electron. Maybe they changed something that made the renderer dependent on the main thread? Or maybe it was changes to scuttlebot replication? Maybe both!

dominictarr:
@mmckegg I added you as an owner on _ssbc_ that means you can now have commit access to everything, and can add other people if necessary!

dominictarr:
also added you as npm owner too

PR #426, opened 2016-10-30
Add missing dependencies to package.json file.

OlegGirko:
This change adds NPM packages that are used but not declared as dependencies
to `dependencies` section of `package.json` file.

Patchwork fails to start without this change.


ahdinosaur:
hey @OlegGirko, thanks!

it looks like these dependencies are missing because of [a recent change](https://github.com/ssbc/patchwork/pull/425) where [`lib/gossip-with-slow-rollout`](https://github.com/mmckegg/patchwork/tree/9f61077cfaca1a7c7145ffe8e5a306fd9a1ca483/lib/gossip-with-slow-rollout) was extracted from [`scuttlebot/plugins/gossip`](https://github.com/ssbc/scuttlebot/blob/master/plugins/gossip/index.js), oops.

looks good @mmckegg

mmckegg:
Ah I see, probably an npm@2 thing! npm3 hides this problem with flat dependencies! Merging!

mmckegg:
Released in v2.12.1 🎉

dominictarr:
these are all dependencies of scuttlebot, so these shouldn't be missing. sometimes, though, npm just doesn't install correctly (!) can you try `rm -rf node_modules` and then `npm install` again, if you get a module error, then post the error here

dominictarr:
oh right, that makes sense

PR #445, opened 2017-03-20
add context menu item to copy message text

ahdinosaur:
the use case is to copy the raw text of a message. if you try to highlight and copy the text from the interface, you will miss all the markdown formatting.

my usual tactic is to use `sbot get {messageId}`, but i wanted to copy a private message's text which is encrypted when displayed with `sbot get`.


mmckegg:
Nice!

PR #446, opened 2017-03-22
redirect https://ssbc.github.io/patchwork to GitHub repo

ahdinosaur:
at the moment the old website is still up with out-of-date info. it seems to have been built from https://github.com/ssbc/ssbc-sitegen/blob/master/repos/patchwork/index.html.js.

this pull request is to change the <https://ssbc.github.io/patchwork> website so it immediately redirects to this repo.

what do people think about this?


clehner:
this means patchwork should be removed from ssbc-sitegen, right? otherwise redirect code should be in ssbc-sitegen. or discard ssbc-sitegen entirely? how to maintain all the gh-pages then?

ahdinosaur:
@clehner i don't know, what do you think is the right way forward?

PR #448, opened 2017-03-25
adds ssb-notifier to enable desktop notification

jekrb:
adds `.use(require('ssb-notifier'))` to server-process.js to enable desktop notifications

### DIFF
--- a/server-process.js
+++ b/server-process.js
@@ -15,6 +15,7 @@ var createSbot = require('scuttlebot')
   .use(require('scuttlebot/plugins/logging'))
   .use(require('./lib/private-with-index'))
   .use(require('./lib/query-with-progress'))
+  .use(require('ssb-notifier'))
 //.use(require('ssb-fulltext')) // disabled for now
 
 module.exports = function (ssbConfig) {


mmckegg:
I think ssb-notifier will have to be forked to work properly with patchwork-next. The alerts will try and open localhost:7777 instead of activating the interface and selecting the correct tab. 

Though, handy hack for anyone who wants notifications now!

jekrb:
@mmckegg cool, no worries! 

@cel and I were just talking about ways we could maybe have ssb-notifier and patchwork/patchwork classic operate on slightly coupled code for now, and then work on way to have ssb-notifier figure out which environment it's running in.

One thing we need to solve is a way of having ssb-notifier work in (for example) an electron app but let it possibly open a notification link in the browser for git ssb web.

I'll begin writing test for ssb-notifier tonight most likely; this might give me some more insight on a good way of achieving the above.

Closing for now :tree:

ahdinosaur:
> One thing we need to solve is a way of having ssb-notifier work in (for example) an electron app but let it possibly open a notification link in the browser for git ssb web.

i had an idea regarding this. what if there was a JSON file in `~/.ssb` (or part of the standard `~/.ssb/config`) that mapped message types to shell commands to open those messages? kinda like file extension handlers / launchers in most operating systems.

so something like

```json
{
  "post": "ssb-patchwork --open ${id}",
  "git-repo": "git ssb open ${id}"
}
```

this of course means implementing the necessary functionality in `patchwork` and `git-ssb-web` in order to open specific messages (and starting the electron app or web server if not already).

mmckegg:
Nice 👍 

Something like that could work well. Also could be used by other apps (such as patchwork or git-ssb) to know how to open certain kinds of links. 

We should probably also think about "protocol handlers" e.g. `ssb://` system wide.

ahdinosaur:
@mmckegg so let's continue this train of thought. if we know how to open messages, then it shouldn't be that hard to write a module `ssb-launcher` which reads this config and runs the appropriate command. then we can all register `ssb-lancher` to handler `ssb://` protocol links, if the links were in format `ssb://${encodeURIComponent(messageId)}`. would that work?

clehner:
i was thinking the message launcher dispatch should be some thing using sbot rpc, but using commands per-type in the config would be more straightforward and approachable now. what should happen when launching a message if the command fails, e.g. opening a git-repo when git-ssb is not installed?

ahdinosaur:
maybe that should be in the config as well, since it seems it could either serve and open a web page or open an electron app to display the error.

PR #449, opened 2017-03-26
window: enable main window to remember size and position

ralphtheninja:
Handles resize, positioning, maximized and full screen. Let me know what you think.


dominictarr:
LGTM

PR #451, opened 2017-03-29
Minor correction to the Pub URL

tiagoantao:
Just the smallest of things, but the URL was not working


ahdinosaur:
hey thanks @tiagoantao!

PR #459, opened 2017-04-06
add missing debian deps for building from source

trqx:
stumbled upon those missing deps when getting from source on x86.


clehner:
merged, thanks

PR #462, opened 2017-04-06
Less whitespace

iffy:
There's an awful lot of whitespace wasting space with white.  Of course design is a matter of opinion, so take this if you want it, but I'm not offended if you don't.


mmckegg:
This seems similar to what @mixmix has done in [patchbay](https://github.com/ssbc/patchbay).

I agree that the whitespace is a bit much right now. Will investigate this at some point. I think some of the problem is that replies take up so much space.

PR #505, opened 2017-04-15
Create LICENSE

sdgathman:
GPL says to include an actual copy of the license.  package.json says GPL-3.0, so here is GPLv3


mmckegg:
Thanks, merged!

PR #513, opened 2017-04-19
use non-hardcoded build script, as in patchbay

dominictarr:
this makes run rebuild use the electron that you have globally installed, and is the same script in patchbay (as provided by @ahdinosaur if I recall)

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
     "test": "echo \"Error: no test specified\" && exit 1",
     "start": "electron index.js",
     "postinstall": "npm run rebuild",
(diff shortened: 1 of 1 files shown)


mmckegg:
@dominictarr @ahdinosaur 

This actually breaks the windows build. We're gonna have to roll this back or use a cross-platform solution

ahdinosaur:
hmm good catch, maybe we want to use [`npm-run-posix-or-windows`](https://www.npmjs.com/package/npm-run-posix-or-windows)?

dominictarr:
oh oops, is there a way to do this that works on windows too? @av8ta

Issue #496, opened 2017-04-11
How to build for windows 


fredhampton:
Are there build instructions for Windows ?

ralphtheninja:
This is a good question. What would it take to build for windows? cc @mmckegg

ahdinosaur:
@ralphtheninja we already do build for windows. :smile: 

instructions here: [%+SBr3lvHbR2yNKASU7FfLX7fTN06yk4ON1qUipaG9fk=.sha256](%25%2bSBr3lvHbR2yNKASU7FfLX7fTN06yk4ON1qUipaG9fk%3d.sha256).

would appreciate if anyone wanted to contribute a "build instructions" page in a `docs` directory and add these instructions to it.

mmckegg:
From %+SBr3lvHbR2yNKASU7FfLX7fTN06yk4ON1qUipaG9fk=.sha256:

> - installed node lts from nodejs.org using msi installer
> - installed git from git-scm.org
> - followed instructions on [node-gyp README](https://github.com/nodejs/node-gyp#installation), option # 1: `npm install --global --production windows-build-tools`
> - configure npm to use windows python executable location: `npm config set python C:\Users\%USERPROFILE%\.windows-build-tools\python27\python`
> - clone down patchwork `git clone https://github.com/ssbc/patchwork` + `cd patchwork` + `npm install` > - `npm start`
>
> THAT IS ALL! WORKED PERFECT!
>
> Also forgot to mention: do all this inside an administrative powershell rather than cmd.

av8ta:
Thanks @mmckegg that's helped me to get further along than before. I still didn't have "make" on my system though so installed mingw-w64. Then added the mingw bin PATH and copied mingw32-make.exe to make.exe in the bin directory. Now make is working but it fails on libtoolize at sodium-prebuilt

> sodium-prebuilt@1.0.22 install C:\code\patchwork\node_modules\sodium-prebuilt
> prebuild --install --preinstall make

./autogen.sh: line 13: libtoolize: command not found
Makefile:8: recipe for target 'configure' failed
make: *** [configure] Error 127
prebuild ERR! build Error: make  failed with exit code 2
prebuild ERR! build     at Object.exports.spawnFailed (C:\code\patchwork\node_modules\prebuild\error.js:31:10)
prebuild ERR! build     at ChildProcess.<anonymous> (C:\code\patchwork\node_modules\prebuild\util.js:116:14)
[...]

PR #514, opened 2017-04-20
add build dependencies for macOS

ahdinosaur:
from https://github.com/ssbc/patchwork/issues/496#issuecomment-294043942 -> https://github.com/mafintosh/node-sodium-prebuilt/issues/10#issue-221242644

thanks @yangwao, @av8ta, @booleanvalue

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,10 +69,19 @@ $ npm install
 $ npm start
 ```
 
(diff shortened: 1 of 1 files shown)


yangwao:
But that are more like scuttlebot dependencies, I think documentation for scuttlebot needs this little trim as well or?

PR #515, opened 2017-04-20
add "Copy External Link" to message context menu

ahdinosaur:
similar to https://github.com/ssbc/patchbay/pull/107

### DIFF
--- a/lib/context-menu-and-spellcheck.js
+++ b/lib/context-menu-and-spellcheck.js
@@ -3,110 +3,124 @@ var {SpellCheckHandler, ContextMenuListener, ContextMenuBuilder} = require('elec
 var {MenuItem, Menu} = remote
 var ref = require('ssb-ref')
 
-window.spellCheckHandler = new SpellCheckHandler()
-window.spellCheckHandler.attachToInput()
+module.exports = setupContextMenuAndSpellCheck
 
-// Start off as US English, America #1 (lol)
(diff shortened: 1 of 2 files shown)


dominictarr:
👍 ideally we need someone else who can build this (for mac) while matt is taking a break.

mmckegg:
@ahdinosaur  This looks cool! Would be nice to use something similar to gateway for external git links too!

https://github.com/ssbc/patchwork/blob/master/modules/app/sync/external-handler/git.js

ahdinosaur:
agree. maybe related to https://github.com/ssbc/patchwork/pull/448#issuecomment-289344832.

mmckegg:
screen shot 2017-04-23 at 1 43 58 pm

This is looking a little busy! So many "Copy ..."

Wonder if there's a way we can tidy this up a bit?

Maybe it should include the name of the viewer you are using to copy?

mmckegg:
Can we put these inside a sub menu?

dominictarr:
c'mon @mmckegg the context menu is already a junk drawer, lets just merge it.

Or, if you really do want to change it, what about something that used depject to pull in things like https://github.com/dominictarr/patch-context/blob/master/defaults/copy-external.js just to make it much easier to add things?

mmckegg:
Okayyy! 🤣

PR #516, opened 2017-04-22
New back button icon and minor header style tweaks

olizilla:
Minor improvements to the header nav styles.

- Fix the default font. (`caption` is a shorthand value for `font` rather than `font-family`)
- New back and forward nav icons, arrow svgs, in the style of chromium.
- Simplify the nav tab buttons.
- Center align the things
- Tweak search input. ~~The existing `input.search` styles were't getting applied. I had to pull them into the base.mcss to get them to apply. If can explain the mcss build process I might be able to fix that.~~ Got it working. Not sure why I need to explicitly mirror the html structure in the css though.

screenshot 2017-04-22 21 10 10

| Before | After |
|-------|-------|
| screenshot 2017-04-22 21 16 45 | screenshot 2017-04-22 21 10 10 |

### DIFF
--- a/main-window.js
+++ b/main-window.js
@@ -59,11 +59,11 @@ module.exports = function (config) {
         h('a', {
           'ev-click': views.goBack,
           classList: [ when(views.canGoBack, '-active') ]
-        }, '<'),
+        }),
         h('a', {
           'ev-click': views.goForward,
           classList: [ when(views.canGoForward, '-active') ]
-        }, '>')
+        })
       ]),
       h('span.nav', [
         tab('Public', '/public'),
--- a/styles/base.mcss
+++ b/styles/base.mcss
@@ -1,7 +1,7 @@
 html, body {
   background: #f5f5f5
   margin: 0
-  font-family: -apple-system, BlinkMacSystemFont, 'avenir next', avenir, 'helvetica neue', helvetica, ubuntu, roboto, noto, 'segoe ui', arial, sans-serif
+  font: caption
   overflow: hidden
   height: 100%
   font-size: 12px
--- a/styles/base.mcss
+++ b/styles/base.mcss
@@ -67,20 +67,6 @@ input {
     padding-right: 12px
     border-radius: 0
   }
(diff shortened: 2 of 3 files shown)


mmckegg:
@olizilla 
> Not sure why I need to explicitly mirror the html structure in the css though.

[mcss uses direct child selectors for everything by default](https://github.com/mmckegg/micro-css#nested-rules-always-use-). You can opt out in specific cases by putting the selector inside `(` parentheses `)`

```
MainWindow {
  input {
    color: black
  }
}
```

translates to `.MainWindow > input { color: black; }`. Whereas,

```
MainWindow {
  (input) {
    color: black
  }
}
```

translates to `.MainWindow input { color: black; }`.

In general I try to avoid using infinite depth child selectors as this can lead to unexpected styling of elements deep in a tree. But the option is there if needed!

mmckegg:
Gosh, we'll need to update the screenshot too!

mixmix:
nice !

dominictarr:
@olizilla awesome!

mmckegg:
patchwork-windows

Maybe we need to think about hiding that double patchwork title on windows and platforms that aren't doing the inset-border.

olizilla:
Glad you all like it!

@mmckegg thanks for the mcss tip, that makes sense. I've been enjoying the basscss / tachyons style of just defining single-purpose, visual css classes, and then combining them in the dom, to dial down the amount of surprises in the cascade. mcss is pretty neat too. I rather like the inline svg handling.

I agree, the double patchwork app title looks wrong. Looks like it'd be simple to hide it for now on none-darwin platforms.

mixmix:
We went mcss with the house that it would make it easier to port / share
components.
Trachyons (or basscss) would perform similarly. I think the other advantage
of mcss is that if you write it first it's kinda like tdd - a prompt to
consciously architect your dom with styling in mind.
So far I've enjoyed it. The direct children thing is a little bit of a
gotcha but i love the specificity

PR #521, opened 2017-04-23
Message ui tweaks

olizilla:
This PR visually simplifies the message and feed-item layout for much calm

- Move card style box shadow to the parent FeedEvent
- Tweak message line-height for readability.
- Fewer border lines. Scuttlebutt knows no borders.
- Tweak padding to visually cluster things.
- Hide repeated `#channel` header in replies.
- Simplify the likes backdown to a regular header meta style... I'd be up for porting the V fingers, from patchwork-classic over to patchwork in another PR if people are up for that. It's rad.

screenshot 2017-04-23 19 30 38

| Before | After |
|--------|------|
| screenshot 2017-04-23 19 24 42 | screenshot 2017-04-23 19 30 38 |

### DIFF
--- a/styles/feed-event.mcss
+++ b/styles/feed-event.mcss
@@ -2,10 +2,10 @@ FeedEvent {
   display: flex
   flex-direction: column
   background: white
-  box-shadow: 0 0 4px 1px rgba(0,0,0,0.1)
+  box-shadow: 0 2px 4px rgba(0,0,0,0.1)
   max-width: 700px
   width: 100%
-  margin: 0 auto 25px
+  margin: 25px auto
 
   -new {
     box-shadow: 0px 0px 2px #ffc800;
@@ -19,11 +19,10 @@ FeedEvent {
   a.full {
     display: block;
     padding: 10px;
-    background: #daecd6;
+    background: f3fafd;
     border-top: 1px solid #bbc9d2;
     border-bottom: 1px solid #bbc9d2;
     text-align: center;
-    color: #759053;
   }
 
   div.replies {
--- a/styles/feed-event.mcss
+++ b/styles/feed-event.mcss
@@ -2,9 +2,10 @@ FeedEvent {
   display: flex
   flex-direction: column
   background: white
+  box-shadow: 0 0 4px 1px rgba(0,0,0,0.1)
   max-width: 700px
   width: 100%
-  margin: 10px auto
+  margin: 0 auto 25px
 
   -new {
     box-shadow: 0px 0px 2px #ffc800;
@@ -37,6 +38,10 @@ FeedEvent {
 
   div.meta {
     font-size: 100%
-    padding: 10px
+    padding: 10px 20px
+    opacity: 0.4
+    :hover {
+      opacity: 0.8
+    }
   }
 }
--- a/styles/message.mcss
+++ b/styles/message.mcss
@@ -1,11 +1,10 @@
 Message {
   display: flex
   flex-direction: column
-  box-shadow: #dadada 1px 2px 8px
-  border: 1px solid #f5f5f5
   background: white
   position: relative
-  font-size: 130%
+  font-size: 120%
+  line-height: 1.4
   flex-shrink: 0
 
   (highlight) {
@@ -54,7 +53,8 @@ Message {
       div.main {
         a.avatar {
           img {
-            width: 35px
+            width: 40px
+            height: 40px
           }
         }
       }
@@ -66,8 +66,7 @@ Message {
   }
 
   header {
-    margin: 15px 15px
-    margin-bottom: 5px;
+    margin: 10px 20px 0
     display: flex
 
     div.mini {
@@ -81,6 +80,7 @@ Message {
       a.avatar {
         img {
           width: 50px
+          height: 50px
         }
       }
 
@@ -133,13 +133,13 @@ Message {
       }
 
       span.likes {
-        color: #ffffff;
-        background: linear-gradient(45deg, #859c88, #87d47d);
-        padding: 5px 8px;
-        border-radius: 10px;
-        display: inline-block;
-        vertical-align: top;
-        margin: -2px 0;
+        margin-right: 10px;
+        color: #286bc3;
+        font-weight: 200;
+        strong {
+          color: #286bc3;
+          font-weight: 200;
+        }
       }
 
       span.private {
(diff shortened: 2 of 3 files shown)


styles/feed-event.mcss
```diff
@@ -37,6 +37,10 @@ FeedEvent {
 
   div.meta {
     font-size: 100%
-    padding: 10px
+    padding: 10px 20px
+    opacity: 0.4
+    :hover {
+      opacity: 0.8
```
mmckegg: Not a huge fan of the way the hover works on this.

Maybe it should be hover on "any" of the message to trigger? Maybe a short transition would help?

Maybe a colour should just be settled on that is a nice compromise that doesn't need a hover at all?

Interesting to note that in the case of replies to recent messages, this heading isn't all that important. It's just telling you why this item has been bumped to the top of your feed.

However in the case that this is an old item, it is actually telling you what the reply is for (it links to the full thread). This case is much more important. Maybe it should be handled slightly different? More like "View Full Thread".

Some thoughts.

mmckegg:
We'll also need to give **new message highlighting** some thought with this new design.

In the old design, new messages get a golden border and a little yellow blip in the meta. If we also use the generated `classList` [here](https://github.com/ssbc/patchwork/blob/master/modules/feed/html/rollup.js#L201), with the new design we get this (which isn't half bad):

screen shot 2017-04-24 at 9 15 36 am

styles/message.mcss
```diff
@@ -201,26 +204,20 @@ Message {
   }
 
   footer {
-    margin: 10px 10px;
-    border-top: 1px solid #eee;
-    padding-top: 10px;
-    margin-top: 5px;
-    padding-left: 5px;
+    margin: 5px 0 20px;
+    padding: 0 20px
```
mmckegg: So looking at this again, and for longer, I can see why the padding is the way it is. Makes sense when there are multiple messages and you want to distinguish which message the actions are for.

Still not quite sure what this means for the case above though (where there's only one message).

styles/feed-event.mcss
mmckegg: Here's an example of replies to an old post (the old post is no longer showing because it's fallen out of the window):

screen shot 2017-04-24 at 9 19 59 am

It's hard to tell in this new design that it is a **reply** and **not a post**.

olizilla:
All good! I'm off to bed now so can take a look tomorrow. If you're in there making things nice then please do feel free to make changes to the PR. I offer it up as a proposal. I've been using it all day and it's easy on the eyes.

mmckegg:
Merged! Ah much cleaner 🎉

I made a few changes though (see master), and I'm very keen for feedback.

- added back in borders between messages (it just felt like there wasn't quite enough to distinguish messages, also the full line cannot be spoofed by a user message)
- made the header of the first message bigger to make it easier to distinguish a post from its replies
- updated about messages to work better with the new styling
- made it so that when you click "x likes" it pops up with a box giving a list of profiles that liked the post (now it makes sense that it looks like a link)

screen shot 2017-04-24 at 12 29 54 pm

And here's the new about messages:

screen shot 2017-04-24 at 12 30 14 pm

mmckegg:
Thanks so much @olizilla!! I really appreciate the time you put in to this 😄 (and seems like a lot of other people do too)

Fresh eyes are great!

So easy to just get used to something. I just see it so much, becomes hard to tell what looks silly 👻

styles/feed-event.mcss
olizilla: > It's hard to tell in this new design that it is a reply and not a post.

I came at this from the "everything's a post" and "all nodes are equal" thoughts, so it seemed nicer to make all posts appear visually equal, rather than make the original poster seem larger than others. 

I think we could find an icon or other visual clues that what is being shown is the middle of a discussion.

As we don't have that yet, I think using icon size is fine for now. We should find a way of visually marking replies as replies; the screenshot above is a good use-case to work on as the original post authors icon isn not visible, so it relies on the user having used the app enough to have a sense of the unseen difference (poster would be larger) rather than being explicit (oh look there is a reply icon).

All of that skips over the point that it says it in the meta... but I'm not totally convinced that the meta adds enough value to justify it's prime position at the top left of the box, but that's a conversation for later ;-)

styles/message.mcss
```diff
@@ -81,12 +85,12 @@ Message {
       a.avatar {
         img {
           width: 50px
+          height: 50px
```
olizilla: I'd go with fixing the `height` in that case, as the design is more sensitive to the vertical requirement of 50px... the width can be whatever mad aspect ratio the user wants... (he says, immediately regretting it...)

olizilla: oh the other reason for fixing both was to reduce the amount off re-layout flicker on load, as on mine it was taking a few seconds to load user avatars. They had a fancy background gradient as a placeholder but I'd only get to see it for a ms as the image loaded as the img tag was rendered with 0px height during the image load.

styles/message.mcss
```diff
@@ -100,6 +104,10 @@ Message {
     }
 
     div.meta {
+      display: flex;
+      flex-direction: column-reverse;
+      align-items: flex-end;
+      justify-content: flex-end;
```
olizilla: The reason for stacking them was just to mirror the name + time combo on the left. I didn't want to make any significant structural changes at this point as I don't know the app well enough and wanted to limit the scope of the PR. 

I'd really like to look at merging the like count with the like button in the actions bar, and adding an icon for it. Do you like the V fingers from patchwork-core? I'd like to see that ported over, but I'm happy to find other ones if no one else likes that.

olizilla:
Yay! I'm glad this got merged and the changes are fine. All my feedback is against a backdrop of "thank you so much for all of this wonderful thing"

> added back in borders between messages (it just felt like there wasn't quite enough to distinguish messages

I respectfully disagree, but they don't offend me too much... I prefer it if threads of posts look connected rather than divided.

> also the full line cannot be spoofed by a user message

I hadn't considered such a thing. Is that a thing? I'd assumed the markdown parser would block arbitrary html tags.

>  made the header of the first message bigger to make it easier to distinguish a post from its replies

No problem... I'd like to make the icon size constant once we've got a visual mechanism / icon for denoting replies in a thread.

> updated about messages to work better with the new styling

👍 ✨ 

> made it so that when you click "x likes" it pops up with a box giving a list of profiles that liked the post (now it makes sense that it looks like a link)

Neat. As mentioned in the inline comments, I'd love to bring back the all in one like and like count button with the V fingers!

mmckegg:
@olizilla 
> I'd love to bring back the all in one like and like count button with the V fingers!

I think if we're gonna change like, we should do it by adding reactions. Then people can "react" however they want. This is already in the protocol. 

Although this will add all sorts of new UX trouble 😆

styles/message.mcss
mmckegg: I'm just gonna leave this how you did it for now. Let's wait for it to break first. Pretty sure all other clients are assuming a square avatar right now anyway.

olizilla:
@mmckegg that sounds fun. If you'd like to jump on a call some time this it'd be good to talk through some of the stuff that's in your head for patchwork futurez, so I can be more use.

ahdinosaur:
> I think if we're gonna change like, we should do it by adding reactions. Then people can "react" however they want. This is already in the protocol.

yes! also i've been thinking multi-reactions would be a great way to signal negative feedback as well.

anyways, thanks @olizilla and @mmckegg for these changes, they look great!

PR #526, opened 2017-04-27
Add custom Window menu with Zoom option

meleyal:
This PR adds the **Zoom** option to the **Window** menu on macOS:

screen_shot_2017-04-27_at_16_53_09

(I think the **Zoom All** and **Bring All to Front** options are there because of the hidden background window where the server process runs).

### DIFF
--- a/index.js
+++ b/index.js
@@ -34,6 +34,16 @@ electron.app.on('ready', () => {
       { type: 'separator' },
       { role: 'togglefullscreen' }
     ]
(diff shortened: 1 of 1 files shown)


ahdinosaur:
thanks @meleyal, loooooks good to me, merging! :cat:

ahdinosaur:
also soz for the 10 day delay, feel free to `@` mention me if you want.

PR #527, opened 2017-04-27
i18n + Kashmiri / Urdu translations

dangerousbeans:


lib/i18n.js
```diff
@@ -0,0 +1,12 @@
+// Multi Language support
+var appRoot = require('app-root-path');
+
+var i18n = require("i18n")
+i18n.configure({
+    locales:['en','ki'],
+    directory: appRoot + '/locales'
+});
+
+// i18n.setLocale('ki')
```
dangerousbeans: Not sure how we handle changing this - I guess could be passed in when making the builds?

mmckegg:
Looks like this PR was folded into #632. Closing!

PR #536, opened 2017-05-17
update level to 1.7.0

ralphtheninja:
This will pull in `leveldown@1.7.0` which contains leveldb version 1.20 and snappy 1.1.4. Also, we have prebuilt versions for electron on linux, osx and windows, see https://github.com/Level/leveldown/releases/tag/v1.7.0 for more info

screenshot from 2017-05-17 17-27-03

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
     "has-network": "0.0.0",
     "insert-css": "~1.0.0",
     "is-visible": "^2.1.1",
-    "level": "~1.6.0",
+    "level": "~1.7.0",
     "level-memview": "0.0.0",
     "map-filter-reduce": "~3.0.3",
     "micro-css": "^2.0.0",


dominictarr:
oh awesome! I think we'll need to update in secure-scuttlebutt too

dominictarr:
hang on, I think this might be wrong, does patchwork even use level directly?

mmckegg:
Only to pin the dep since 1.5 was horribly broken on some Linux versions.

ralphtheninja:
> oh awesome! I think we'll need to update in secure-scuttlebutt too

Ah ok, so we should remove level from package.json and only update secure-scuttlebutt?

mmckegg:
I don't trust `level` minor releases, hence the pinned `~` reference in Patchwork.

mmckegg:
Gosh I'm suddenly feeling very ranty about npm dependency management 😢 

But anyway, I merged this. Doesn't feel like a very nice solution to this problem though.

dominictarr:
btw, secure-scuttlebutt is updated to level 1.7 so when we merge flume stuff (very shortly) i'll all be good.

PR #537, opened 2017-05-18
Adds patch-gatherings

pietgeursen:


dominictarr:
question: what is stopping this from just working with patch-gatherings?

ahdinosaur:
> question: what is stopping this from just working with patch-gatherings?

coordinating on common interfaces (APIs) is _hard_, so there are still some differences between how `patchwork` and `patchbay` implement some of the more nuanced behavior, as [you discovered with things like the message composer](https://github.com/ssbc/patchbay/pull/114), or page routing, or how streams of messages are "rolled-up" (or not), or how styles are done, or how the context menu is done, etc.

so my understanding is that `patch-gatherings` needs a common interface between `patchwork` and `patchbay` in order to _just work_, so at the moment it is has separate external interfaces for each: [`patchwork-gatherings`](https://github.com/pietgeursen/patchwork-gatherings) and [`patchbay-gatherings`](https://github.com/pietgeursen/patchbay-gatherings).

although this is only my high-level perspective as an observer, i'm sure @pietgeursen @mixmix will be able to do much better at describing the specific common interfaces we need in order to support `patch-gatherings` in both clients.

mixmix:
Is plugged in and kinda works

The client specific stuff is things like "how does this thing render
messages" (different depject api currently), and "how does this thing let
you get to gatherings"... currently people need to type /gatherings in,
ideally there is a button somewhere though.

Finally, matt has slightly differently styles which means the styles from
patch-gatherings 80% work... Things like the width are wrong, and the
assumption about where avatar images go.

Lots of little "not quite modular enough"

dominictarr:
Yes I know it's hard. If we don't put effort into stuff like this we'll be unable to add features. Afterall, patchwork@2 died because no one wanted to maintain it. It's either that or periodically rewrite from scratch.

@mixmix is it that the width is just different? or that it is expressed via a different mechanism? can you describe (or screenshot) the effects of avatar placement?

mixmix:
this is working now. MVP.
Keen to merge and make builds, and also have conversations about what was and is hard about developing a patch-plugin across both apps.

I anticipate there's going to be a bunch of learning and iteration as we proceed.

Things I've noticed already:  
- bay and work have diverging dom structure for message
- bay and work have different approaches to folding in styles, which is a bit of a pain (no depject)
- spinning things up in work and bay simultaneously is too hard because of the different sbots.

Anyway we don't need all of this solved at once @mmckegg if you're happy enough and want to throw us a a :+1: we can start making events we can invite you to ... and finally cleave you from facebook :heart:

ahdinosaur:
has anyone else noticed this change affecting Patchwork performance? so far after upgrading to this version i haven't been able to use Patchwork beyond the initial load because it hangs at 100% cpu. maybe completely unrelated.

mixmix:
That don't sound good. I've not had an issue, but it's not my primary
client. Can you confirm Matt?

mixmix:
cc @mmckegg ^^

I've written up some release notes : https://github.com/ssbc/patchwork/releases

mmckegg:
Looks good!

My only tiny nitpick would be that I wouldn't mention the level update as this doesn't actually affect the release in any way.

mixmix:
yeah i didn't know if it would and thought that if it did having this line
might save someone a bunch of time.

PR #538, opened 2017-05-20
disable attachments on private threads

dominictarr:
people (reasonably) assume they are encrypted, but they arn't encrypted currently.

https://github.com/ssbc/patchbay/pull/114

### DIFF
--- a/modules/message/html/compose.js
+++ b/modules/message/html/compose.js
@@ -21,7 +21,8 @@ exports.needs = nest({
 exports.gives = nest('message.html.compose')
 
 exports.create = function (api) {
-  return nest('message.html.compose', function ({shrink = true, meta, prepublish, placeholder = 'Write a message'}, cb) {
+  return nest('message.html.compose', function (opts = {}, cb) {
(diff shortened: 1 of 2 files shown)


dominictarr:
so I think we should disable them until we have them be encrypted properly.

modules/message/html/compose.js
```diff
@@ -21,7 +21,8 @@ exports.needs = nest({
 exports.gives = nest('message.html.compose')
 
 exports.create = function (api) {
-  return nest('message.html.compose', function ({shrink = true, meta, prepublish, placeholder = 'Write a message'}, cb) {
+  return nest('message.html.compose', opts = {}, function (, cb) {
```
ahdinosaur: i think this line should be

```js
return nest('message.html.compose', function (opts, cb) {
```

dominictarr: yeah, sorry I pushed a new commit that fixes that.

dominictarr:
that would be awesome!

dominictarr:
definitely worth a :tada:

PR #541, opened 2017-05-24
Use key instead of code

### LINKED ISSUE #534: Number pad enter key does not work in search bar
If you enter a search term in the search bar, pressing the number pad enter key has no effect. I expected the behavior to be the same as when I press the standard enter key.

dmamills:
Small change to fix #534


ahdinosaur:
hey, thanks @dmamills!

Issue #528, opened 2017-04-28
Show number of pending updates in Patchwork icon on Mac


mitar:
Many applications show a number badge on their icon for number of updates, notifications, etc. Patchwork could do the same on Mac.

dmamills:
This looks possible using [electron's `setBadge`](https://electron.atom.io/docs/api/app/#appdocksetbadgetext-macos)

mmckegg:
This was added in b1b0ee71c45997c433aba9eaf0175aa177983a29

PR #544, opened 2017-05-29
Show Pending Updates on Dock

dmamills:
As suggested in #528. Any feedback is welcomed :)

### DIFF
--- a/lib/window.js
+++ b/lib/window.js
@@ -50,7 +50,10 @@ module.exports = function Window (config, path, opts) {
 
   window.on('closed', function () {
     electron.ipcMain.removeListener('ready-to-show', handleReadyToShow)
+    electron.ipcMain.removeListener('update-badge-count', updateBadgeCount)
   })
+
+  electron.ipcMain.on('update-badge-count', updateBadgeCount);
 
   window.loadURL('file://' + Path.join(__dirname, '..', 'assets', 'base.html'))
   return window
@@ -62,5 +65,9 @@ module.exports = function Window (config, path, opts) {
       window.show()
       electron.ipcMain.removeListener('ready-to-show', handleReadyToShow)
     }
+  }
+
+  function updateBadgeCount(ev, count) {
+    electron.app.setBadgeCount(count);
   }
 }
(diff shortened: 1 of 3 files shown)


JesseObrien:
👍

mmckegg:
More than one view uses rollup.js so the different counts will override each other. Not sure if this is the best behaviour.

dmamills:
I didn't notice it was used multiple times thanks. I thought it would be easy to keep it close to the `updates` observables as it made dismissing the badges nice and simple. Is the best way to go to create a dedicated stream from `api.sbot.pull.feed` and just count it up myself?

mmckegg:
Thanks for tackling this! Unfortunately, it's probably a harder problem than it initially appears.

We need to decide exactly what the "pending updates" should reflect. 

- Should it be messages in any of the tabs? 
- Or just public? 
- Or just mentions? 
- Maybe it should be the current active tab?

What if the main window is closed? On Mac, sbot keeps running in the background. 

You could sit on`pull.feed` but then you need to manually figure out which updates are relevant to the current user. There is a lot of "FOAF" data in there.

Maybe we should choose something that is good enough for now, and then once #509 lands, we might be able to do it better.

My suggestion **for now** would be to add up the pending counts for all tabs and display that (or at least public + private). You could do it with something like this in main window:

```js
var electron = require('electron')
var computed = require('mutant/computed')
var watch = require('mutant/watch')

/////

var pendingCount = computed([
  views.get('/public').pendingCount,
  views.get('/private').pendingCount
], (...counts) => {
  return counts.reduce((a, b) => a + b)
})

watch(pendingCount, count => {
  electron.remote.app.setBadgeCount(count)
})
```

mmckegg:
In regards to **"On Mac, sbot keeps running in the background"**:

Maybe a good trick would be to just hide the main window if it gets closed (rather than destroying it), that way updates will keep working in the background with the code above. The downside is, it would use more cpu and ram (well the same amount as if you had it open). But this is the way that most electron apps (like WebTorrent) handle backgrounding.

mmckegg:
Sorry, should be `views.get('/public').pendingUpdates` not `pendingCount`

dmamills:
Thanks for the super helpful reply @mmckegg! The point in the right direction much appreciated. I'm going to give this another go later on today :)

dmamills:
Worked like a charm 👍

Issue #476, opened 2017-04-08
add browse channels page


mmckegg:
List channels that people in your network subscribe to.

ahdinosaur:
solved with https://github.com/ssbc/patchwork/pull/547

PR #547, opened 2017-06-02
Add Channels Page (#476)

peterbraden:
Channels sidebar links to a new page with all channels in.

### DIFF
--- a/modules/page/html/render/public.js
+++ b/modules/page/html/render/public.js
@@ -135,7 +135,11 @@ exports.create = function (api) {
         h('button -pub -full', {
           'ev-click': api.invite.sheet
         }, '+ Join Pub'),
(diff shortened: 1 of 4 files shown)


modules/page/html/render/public.js
```diff
@@ -135,7 +135,11 @@ exports.create = function (api) {
         h('button -pub -full', {
           'ev-click': api.invite.sheet
         }, '+ Join Pub'),
-        when(computed(channels, x => x.length), h('h2', 'Active Channels')),
+        when(computed(channels, x => x.length),
+             h('h2',
```
mmckegg: It's a little bit strange having this as a hyperlink on the title as this is not a common pattern anywhere else in patchwork.

I would prefer if this were a seperate button at the bottom of the channel list "More Channels..."

peterbraden:
fixed

PR #587, opened 2017-07-06
Start ssb-notifier if installed

lancew:
As per git-ssb start ssb-notifier if it is installed.

Idea here being to replicate the behavior of git-ssb and autostart as background process the notifier.

I have had it running like this for a week or so. At this stage with two additional services potentially starting automatically in the background (git-ssb and ssb-notifier) perhaps a UI menu where they can be enabled/disabled is needed.

I get few messages so notifications for me are infrequent, more busy users may find the notifications intrusive perhaps? So this patch perhaps not desired in grander scheme of the project?

Consider this perhaps a discussion starter?

### DIFF
--- a/server-process.js
+++ b/server-process.js
@@ -44,4 +44,15 @@ module.exports = function (ssbConfig) {
   process.on('exit', () => {
     gitSsb.kill()
   })
+  
+   // attempt to run ssb-notifier if it is installed and in path
+  var ssbNotifier = spawn('ssb-notifier', {
+    stdio: 'inherit'
+  })
+  ssbNotifier.on('error', () => {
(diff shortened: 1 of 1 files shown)


mmckegg:
As mentioned [here](https://github.com/ssbc/patchwork/pull/448), it would be nice if `ssb-notifier` was updated to be able to open in Patchwork if clicked. There would be a bit of work to do on both sides. Once this is addressed, I would be more open to merging this change. 

What I would prefer to do is roll something directly into patchwork that could be configured via settings. 

Closing this PR. Continue discussion in #633

lancew:
Taking a look at #633. :-)

PR #602, opened 2017-08-07
only count mentions for badge count

dangerousbeans:
I feel the red badge over the icon in the tray should show updates when they're relevant to you, not just any new updates anywhere on the network not yet viewed.

Just counting mentions should include private messages also


mmckegg:
Yeah, you're probably right! Would be more useful this way.

Gonna merge.

PR #618, opened 2017-09-09
Theme picker

mixmix:
Hey @alanshaw here's a them picker which doesn't even need a restart ! 

This first pass has the themes seperated (in full duplicate) in `/styles/light` and `/styles/dark`

This is in anticipation of maybe wanting to fully extract these themes. I had a think about whether we could also support themes which behave more like a 'patch' and just over-ride the default theme. I'm open to that, as well, it may be a bit easier to develop. What do you think?


mixmix:
If this feels useful, I'd love to hand this over to you @alanshaw - as in feel free to do what you want with it. Happy to discuss implementation more.

main-window.js
```diff
@@ -65,7 +67,15 @@ module.exports = function (config) {
     electron.remote.app.setBadgeCount(count)
   })
 
-  insertCss(require('./styles'))
+
+  watch(api.settings.obs.get('patchwork.theme', 'light'), name => {
+    Array.from(document.head.children)
```
ahdinosaur: is this to reset the previous styles? i reckon would make more sense to have a single `<style>` element with an id, then override just the element. is easy with https://github.com/bendrucker/insert-styles

mixmix: good to know about that. I need to keep building and polishing other things. 

I'm really feeling some of the provocation of that talk "worse is better" at the moment - just make it bad, and work on the important things. For me, the important part here was enabling Alan, and making this tidy enough for people to experiment with. It could be a bad idea, or it could be something which should be branched out in the future, I don't know. 

If you'd like to improve it, I'm totally happy for others to change / adapt what I've done here.

If you think this is too light an involvement to be net good, would love to hear :heart:

styles/index.js
```diff
@@ -1,20 +1,4 @@
...
-  }
-})
-
-const flatpickr = require.resolve('flatpickr/dist/flatpickr.css')
-additional += fs.readFileSync(flatpickr) + '\n'
-
-module.exports = compile(result) + additional
+module.exports = {
+  light: require('./light'),
+  dark: require('./dark')
```
alanshaw: I think it's worth getting this code merged asap. Since all the CSS files have changed location this branch is quickly going to become conflicty with other branches that need to change the CSS.

alanshaw:
This is so awesome! ✨ 

What's the best path to getting this merged? Some options:

1. We can merge as is - light/dark button has no visual effect yet
2. Remove dark folder from styles and merge then - weird UI with only one button
3. Apply work from https://github.com/ssbc/patchwork/tree/dark to this branch and wait till complete before merging
4. Apply work from https://github.com/ssbc/patchwork/tree/dark to this branch, merge it and separate PR to finish styles later (WIP dark styles)

I'm more keen to get this merged sooner, since all the style files have changed location this branch will get conflicty quite quickly if others are changing CSS.

styles/index.js
mixmix: merged into your ahhh dark branch right?

if yes then go for it

mmckegg:
Awesome! 😄

mixmix:
@alanshaw others shouldn't be changing styles much at the moment.

This setup should show you visible changes when you hit those buttons. If it doesn't something is wrong.

I merged the dark branch in ... well ... I did `git checkout origin/dark styles` and then moved the styles into the dark folder because when i tried merging dark in it over-wrote my light styles...

mixmix:
ok here's what I've done: 
- created and merged https://github.com/ssbc/patchwork/pull/619
  - adds settings page
  - moves current style out into 'light' style
- pulled all the dark styles into a new PR that is aimed at https://github.com/ssbc/patchwork/pull/621

This is in master now

regrets :sob:  
- i ... don't look at theme-picker, or this PR any more... i did some git things I don't want to talk about

alanshaw:
:thumbsup:

PR #624, opened 2017-09-13
implement attachment block for files >5MB

mixmix:
selection_387

This is a straight port from my patchbay work. (apart from the hot pink, which is more legible)

Pretty simple

### DIFF
--- a/modules/message/html/compose.js
+++ b/modules/message/html/compose.js
@@ -53,7 +53,26 @@ exports.create = function (api) {
       placeholder
     })
 
+    var warningMessage = Value(null)
+    var warning = h('section.warning',
+      { className: when(warningMessage, '-open', '-closed') },
+      [
+        h('div.warning', warningMessage),
+        h('div.close', { 'ev-click': () => warningMessage.set(null) }, 'x')
+      ]
+    )
     var fileInput = api.blob.html.input(file => {
+      const megabytes = file.size / 1024 / 1024
+      if (megabytes >= 5) {
+        const rounded = Math.floor(megabytes*100)/100
+        warningMessage.set([
+          h('i.fa.fa-exclamation-triangle'),
+          h('strong', file.name),
+          ` is ${rounded}MB - the current limit is 5MB`
+        ])
+        return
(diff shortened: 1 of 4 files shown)


modules/message/html/compose.js
```diff
@@ -53,7 +53,26 @@ exports.create = function (api) {
...
+    )
     var fileInput = api.blob.html.input(file => {
+      const megabytes = file.size / 1024 / 1024
+      if (megabytes >= 5) {
+        const rounded = Math.floor(megabytes*100)/100
+        warningMessage.set([
+          h('i.fa.fa-exclamation-triangle'),
+          h('strong', file.name),
+          ` is ${rounded}MB - the current limit is 5MB`
+        ])
```
mixmix: Ideally we'd get sbot to rm that file ... https://scuttlebot.io/apis/scuttlebot/blobs.html#rm-async

modules/message/html/compose.js
```diff
@@ -53,7 +53,26 @@ exports.create = function (api) {
...
+    var warningMessage = Value(null)
+    var warning = h('section.warning',
+      { className: when(warningMessage, '-open', '-closed') },
+      [
+        h('div.warning', warningMessage),
+        h('div.close', { 'ev-click': () => warningMessage.set(null) }, 'x')
+      ]
+    )
     var fileInput = api.blob.html.input(file => {
+      const megabytes = file.size / 1024 / 1024
```
ahdinosaur: this should probably be a function `bytesToMegabytes`

modules/message/html/compose.js
```diff
@@ -53,7 +53,26 @@ exports.create = function (api) {
...
+      if (megabytes >= 5) {
+        const rounded = Math.floor(megabytes*100)/100
```
ahdinosaur: i'm confused what this does. does this round a decimal number to two digits?

mixmix: yep, is there a tidier way ?

mixmix:
Good enough to merge ?

PR #626, opened 2017-09-18
Add setting for filtering following messages

mlegore:
I have found that follow messages to be distracting in the feed, for me anyway they make up more than half of the content, so I added a setting to allow filtering of those messages. I think there might be a better way to hook up the observables in rollup.js, I'm open to suggestions for it.

### DIFF
--- a/modules/feed/html/rollup.js
+++ b/modules/feed/html/rollup.js
@@ -27,7 +27,8 @@ exports.needs = nest({
   'message.sync.root': 'first',
   'feed.pull.rollup': 'first',
   'sbot.async.get': 'first',
-  'keys.sync.id': 'first'
+  'keys.sync.id': 'first',
+  'settings.obs.get': 'first',
 })
 
 exports.gives = nest({
@@ -77,6 +78,13 @@ exports.create = function (api) {
         when(loading, h('Loading -large'))
       ])
     ])
+
+    const filters = api.settings.obs.get('filters')
+    rootFilter = getFilter(filters())
+    filters((filterSettings) => {
+      rootFilter = getFilter(filterSettings)
+      refresh()
+    })
 
     onceTrue(waitFor, () => {
       // display pending updates
@@ -314,6 +322,12 @@ function getType (msg) {
 
 function returnTrue () {
   return true
+}
+
+function getFilter (filterSettings) {
+  return function(msg) {
+    return !(filterSettings.following && getType(msg) === 'contact')
+  }
 }
 
 function byAssertedTime (a, b) {
--- a/modules/feed/html/rollup.js
+++ b/modules/feed/html/rollup.js
@@ -27,8 +27,7 @@ exports.needs = nest({
   'message.sync.root': 'first',
   'feed.pull.rollup': 'first',
   'sbot.async.get': 'first',
-  'keys.sync.id': 'first',
-  'settings.obs.get': 'first',
+  'keys.sync.id': 'first'
 })
 
 exports.gives = nest({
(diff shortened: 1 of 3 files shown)


modules/page/html/render/settings.js
```diff
@@ -44,6 +45,23 @@ exports.create = function (api) {
...
+
+              return [
+                h('button', {
+                  'ev-click': () => api.settings.sync.set({
+                    filters: {following: !filter}
+                  }),
+                  style
+                }, filter ? 'Hide following messages' : 'Show following messages')
+              ]
+            })
```
mmckegg: This interface is a bit confusing here because it is not clear if the state on the button "Hide following messages" is the current one or what will be set if clicked. This might be better as a checkbox:

```js
h('label', [
  h('input', {
    type: 'checkbox', 
    checked: filterFollowing, 
    'ev-change': (ev) => api.settings.sync.set({
      filters: {following: ev.target.checked}
    })
  }), ' Hide following messages'
])
```

mlegore: Yes, that is a lot more clear that way.

modules/feed/html/rollup.js
```diff
@@ -78,6 +79,13 @@ exports.create = function (api) {
       ])
     ])
 
+    const filters = api.settings.obs.get('filters')
+    rootFilter = getFilter(filters())
+    filters((filterSettings) => {
+      rootFilter = getFilter(filterSettings)
```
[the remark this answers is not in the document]
mlegore: Ah, still getting used to this style of dependency injection while looking through the code.

The only problem I have with this is I can't seem to get the UI to re-render and filter out following messages that have already been seen. As you scroll down, new messages are filtered, but messages that have already been rendered aren't hidden until I manually refresh the page. 

I was thinking I can either pass in the observable into rollup.js and attach a callback to that which refreshes rollup, or maybe add an observable filter and AND it with the rootFilter. I'm thinking I'll go with the latter.

modules/page/html/render/public.js
```diff
@@ -78,9 +80,14 @@ exports.create = function (api) {
...
       },
+      observableFilter: computed([filters], function(filterFollowing) {
+        return {
+          filter: (msg) => !(filterFollowing && getType(msg) === 'contact')
+        } // Put this observable fn in object so it's not flattened
+      }),
       waitFor: computed([
         following.sync,
-        subscribedChannels.sync
+        subscribedChannels.sync,
```
mlegore: Unnecessary comma..

modules/page/html/render/public.js
```diff
@@ -78,9 +80,14 @@ exports.create = function (api) {
           return isSubscribed || id === author || following().has(author)
         }
       },
+      observableFilter: computed([filters], function(filterFollowing) {
+        return {
+          filter: (msg) => !(filterFollowing && getType(msg) === 'contact')
```
mmckegg: You can also add the `passthru` option to prevent flattening:

```js
computed([filters], ..., {
  passthru: true
})
```

modules/feed/html/rollup.js
```diff
@@ -41,7 +41,8 @@ exports.create = function (api) {
     bumpFilter = returnTrue,
     displayFilter = returnTrue,
     updateStream, // override the stream used for realtime updates
-    waitFor = true
+    waitFor = true,
+    observableFilter // filter that can change during runtime
```
mmckegg: I don't think this is necessary for the automatic refresh.

You could call `feedView.reload()` inside _public.js_.

Based on my original example:

```js
var filters = api.settings.obs.get('filters')

var feedView = api.feed.html.rollup(getStream, {
  rootFilter: function (item) {
    if (!filters()) return true
    return !(filters().following && getType(msg) === 'contact')
  },
  ...
})

// call reload whenever filters changes (equivalent to the refresh from inside rollup)
filters(feedView.reload)
```

modules/page/html/render/public.js
mlegore: I overengineered this here, using your suggestion for public.js calling refresh I don't need this at all. Much better!

mmckegg:
Cool, merged!

mlegore:
Sweet, my first open source PR!

PR #629, opened 2017-09-24
Travis CI

harlantwood:
Even without tests, I think CI is pretty valuable here, to make sure everything at least builds correctly.


dominictarr:
:+1: we should probably have ssbc/patchwork-builder too

harlantwood:
Great, shall we merge this?

PR #632, opened 2017-09-29
I18n reload (It's serious this time :) )

gmarcos87:
This pull request tries to unify all the porpose that is around i18n and patchwork.
Basically it proposes:
- [x] I18n via depject
- [x] i18n-js as a translation engine
- [x] Human-time wrapping
- [x] Language selection in the setup screen
- [x] Configuration stored via settings.sync.set
- [x] If there is no language configuration, try autodetection
- [x] Auto-loading of all languages
- [x] Rerender all screens when detecting language change


lib/window.js
```diff
@@ -19,7 +19,7 @@ module.exports = function Window (config, path, opts) {
 
       var config = ${JSON.stringify(config)}
       var data = ${JSON.stringify(opts.data)}
-      var title = ${JSON.stringify(opts.title || 'Patchwork')}
+      var title = ${JSON.stringify(opts.title || "Patchwork" )}
```
mmckegg: Is there a specific reason for switching to use double quotes on some strings?

Patchwork code mostly conforms to [JavaScript Standard Style](https://standardjs.com/rules.html) (and I personally use this as a linter when working on it), and this only allows double quotes (") when you want to escape single ones.

modules/page/html/render/settings.js
```diff
@@ -30,29 +36,51 @@ exports.create = function (api) {
...
-            h('h2', 'Theme'),
-            h('select', {
-              style: {
-                'font-size': '120%'
-              },
-              value: currentTheme,
-              'ev-change': (ev) => api.settings.sync.set({
-                patchwork: {theme: ev.target.value}
+            h('h2', i18n('Theme')),
+            computed(currentTheme, currentTheme => {
```
mmckegg: When you rebase/merge with the latest master, you'll see that the theme picker has been [replaced with a `select` dropdown box](https://github.com/ssbc/patchwork/commit/df3a0af70b34b408b510eafb21126103772656a3) instead of the toggle buttons. I suggest you use that for the language picker too!

mmckegg: Oh wait, I see that this has already merged in the change. But it has overridden it with the old code putting the original theme picker back in.

Can you update this to use the new code from https://github.com/ssbc/patchwork/commit/df3a0af70b34b408b510eafb21126103772656a3 ?

plugs/intl/sync/i18n.js
```diff
@@ -0,0 +1,99 @@
...
+
+  //Get translation  
+  function i18n (value) {
+    _init()
+    return i18nL.__(value)
+  }
+
+  function time (date){
+    return date
+      .replace(/from now/, i18n('form now'))
```
mmckegg: just noticed a typo on this string: should be "from now" not "form now"

mmckegg: It also has matching typos in the string tables.

mmckegg:
Okay, I've gone ahead and fixed the things I mentioned above and merged!

Also I think I'm starting to understand how the i18n module works 😃 

Looks like there is a lot we can do later to improve the quality of the translations including substitutions (word ordering) and pluralisation.

gmarcos87:
I also never did an i18n implementation on such large projects. I encountered several challenges, many of them unresolved.
In Spanish, as in many other languages, words are assigned an implicit gender. Usually the generalizations are made in masculine form, but seen from a more critical point of view this reinforces the preexisting inequalities between genres. This forces to determine the gender of the recipient, or impose the gender definition of the recipient. I do not know if you understand what I mean, but it is a problematic situation that is usually ignored, or accepted as a final determination.

I also noticed two more technical elements to solve, the order of the variables in the sentences and how to incorporate future translations that come from external modules.

For the order of the variables it may be better to opt for another translation engine, i18nline has support for pluralization and multiple variables.

I do not know if I could incorporate external translations, my knowledge of depject does not go so far. It would be interesting to find a solution, even giving the user the ability to change the translation for what seems most convenient and share it on your network.

PR #636, opened 2017-10-02
started german locale

cryptix:
Started a basic german locale. Definitely incomplete and has typos but it's something to start with.

Open questions would be how to deal with terms like _Like_, _Thread_,  _Gathering_, _Channel_ that don't have good translations. I'd keep them English.

ps: found that _Reply_ wasn't using l18n yet.


locales/de.json
```diff
@@ -0,0 +1,144 @@
+{
+	"$name": "Deutsch (German)",
```
mmckegg: You can add `de: "German"` to the english translation (and others) to handle the "(German)" part automatically depending on the currently selected language

overrides/patchcore/message/html/action/reply.js
```diff
@@ -1,10 +1,13 @@
...
 
+exports.needs = nest({'intl.sync.i18n': 'first'})
+
 exports.gives = nest('message.html.action')
 
 exports.create = (api) => {
+  const i18n = api.intl.sync.i18n
   return nest('message.html.action', function reply (msg) {
-    return h('a', { href: msg.key, anchor: 'reply' }, 'Reply')
+    return h('a', { href: msg.key, anchor: 'reply' }, i18n('Reply'))
```
mmckegg: Thanks, good catch!

locales/de.json
cryptix: Ah, I see! I wonder how that worked.

cryptix:
hmm.. thanks but the collab access somehow didn't take?

> Only those with write access to this repository can merge pull requests.

mmckegg:
@cryptix did you accept the email invite?

cryptix:
sorry about the noise. found it and have the merge button now.

cryptix:
Some of these are a bit lax and clunky but I have to use it a bit to see how it feels.

PR #642, opened 2017-10-07
Don't register channel subscription / unsubscription / vote messages as 'recent activity'

Happy0:
Often I click one of the channels that are 'active' to the left of the patchwork client, only to discover the only activity was a subscription / unsubscription message.

I would prefer if a channel was only considered 'active' when there's been a recent message. I'm not sure how many other people have that opinion.

Is this a change you'd be interested in?

This needs more testing, but I thought I'd start a pull request early to make sure it's something you're interested in having first :).

### DIFF
--- a/sbot/channels.js
+++ b/sbot/channels.js
@@ -5,6 +5,13 @@ module.exports = function (ssb, config) {
   return ssb._flumeUse('patchwork-channels', FlumeReduce(1, reduce, map))
 }
 
+function isActivityMessage(msg) {
+  var isVote = msg.value.content.type === "vote";
+
+  return !isVote && msg.value.content.subscribed !== false
+    && msg.value.content.subscribed !== true;
+}
+
 function reduce (result, item) {
   if (!result) result = {}
   if (item) {
@@ -14,7 +21,11 @@ function reduce (result, item) {
         value = result[channel] = {count: 0, timestamp: 0}
       }
       value.count += 1
-      if (item[channel].timestamp > value.timestamp) {
+
+      // We don't update the timestamp if the messsage was just somebody subscribing
+      // or unsubscribing from the channel, or it is a vote as we don't want it to register as
+      // 'recent activity'.
+      if (item[channel].isActivityMessage && item[channel].timestamp > value.timestamp) {
         value.timestamp = item[channel].timestamp
       }
     }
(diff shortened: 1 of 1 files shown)


cryptix:
Very much in favor of this and the change itself is small also looks sound.

cryptix:
hmm..

i tried the branch and subscribed to a new channel which then jumped to the top of my  feed. I thought it shouldn't do that? will try to debug it later.

Happy0:
I think when you subscribe to a new channel it might be a different code path from somebody else subscribing to it. It's certainly distinguished in the UI by the green bar at the left.

I'll see what I can find about this later when I get home too :).

I've been able to see it **not** updating the list by adding print statements about whether it thinks something should update the 'active' timestamp or not (especially at startup when catching up with a bunch of missed messages.)

One thing I'm not 100% sure about is whether you still get channels being bumped on the left when you happen to be using the client as someone votes / subscribes. I don't know if that triggers a different code path or something or whether my eyes were deceiving me. This is also something I need to investigate more.

mmckegg:
This is a good approach @Happy0.

The reason that I did not implement this was I thought it could be a cool way to find out about new channels that you don't yet subscribe to. But now that I think about it, just seeing this in the feed is probably good enough.

I'm going to go ahead and merge this.

Happy0:
Nice :D

mmckegg:
Oh with this approach, you'd also need to modify https://github.com/Happy0/patchwork/blob/abb85fa0947760cd74602bc9f246265ae015e3e1/plugs/channel/obs/recent.js#L42 to filter out `isActivityMessage`.

I think another method to do this a little more simply would be to just filter out the activity messages all together. They don't need to be in here.

```js
function map (msg) {
  if (msg.value.content && !isActivityMessage(msg)) {
    var channel = normalizeChannel(msg.value.content.channel)
    if (channel) {
      return {
        [channel]: {timestamp: msg.timestamp}
      }
    }
  }
}
```

I'm going to implement it like this.

Happy0:
Good point. I actually looked for a 'filter' when I started doing it, but I didn't realise you could just return null from the map and check for it in the reduce function (which I see is actually already the case.)

Anyway, thanks for the merge. It's nice to have s commit in the patchwork git log :D

PR #648, opened 2017-10-14
Do not display 'show <n> updates' message for messages patchwork cannot render

Happy0:
Sometimes I've used patchwork at the same time as patchbay while playing chess. I've noticed that when chess moves are made, patchwork will show a 'Show <n> updates' on the main public feed. However, when you click it, there are in fact no new messages shown as patchbay cannot render the new messages. This is also the case for other types of messages that aren't ssb-chess related.

I propose a change where the 'Show <n> updates' message is only shown if they are updates that patchwork can render.

Is this a change you'd like?

If so, would you like me to change my strategy for discovering if a message is renderable by patchwork (and hence worth adding to the count of unseen messages)?


Happy0:
It looks like there's one or more files that 'give' `message.html.render` , so I'd have to implement `message.html.canRender` for those in patchcore too.

Before doing that, I'll wait until I hear if you'd rather I used another approach or not.

mmckegg:
@Happy0 

While this solution does seem a little complicated on the surface (and I suspect it shows up inadequacies in the way we structured the depject stuff), I can't think of any other way to handle this.

I think this is the best approach given the way the app currently works. Keep going, keen to merge!

plugs/message/html/render/about.js
```diff
@@ -17,67 +17,80 @@ exports.needs = nest({
...
-        layout: 'mini'
-      }, opts)), { msg }))
+      return elements
     }
-
-    return elements
   })
+
+  function isRenderable(msg) {
+    if (msg.value.content.type !== 'about') return false
```
mmckegg: @Happy0 Something you might want to consider is that if you return `false` from a depject `"first"`, that will be final answer. You need to return `undefined` or nothing at all if you want it to continue to the next one.

Happy0: Thanks! :)

Happy0:
@mmckegg I have created a pull request on patchcore and returned 'undefined' rather than 'false' from my `isRenderable` functions in this pull request.

I need to go to bed at the moment. I'll give my changes more testing when I have time over the next couple of days :)

package.json
```diff
@@ -40,7 +40,7 @@
     "mutant-pull-reduce": "^1.1.0",
     "obv": "0.0.1",
     "patch-settings": "^1.0.1",
-    "patchcore": "~1.13.2",
+    "patchcore": "~1.13.3",
```
mmckegg: What I normally do in this case is reference the git repo and commit directly in the package json.

If you do `npm install Happy0/patchcore#canRender --save` it will update the entry in the package.json to:

```
"patchcore": "github:Happy0/patchcore#canRender"
``` 

and will use the version directly off github. The only downside is that if you delete the branch, you can no longer install that particular fork.

Happy0: Oh, cool! I didn't know you could do that :). Done.

Goodnight!

Happy0:
So far I've tested:
* New posts show up on the 'show <n> messages' button
* chess messages do not
* Gatherings, profiles, friends /following / followers / (patchcore) posts still render.
* (patchcore) subscribed / unsubscribed from channel renders in channels.
 
Still needs tested:

* (patchcore) github issues still render when using my patchcore branch.
* Find out where html.render.vote in patchcore is used, and test it still renders.

Anything I've missed?

I'm going to work just now. Will test the other stuff later :).

Happy0:
I tested that github issues and votes still render when running `npm start` in patchcore (with npm start)

If you want me to test anything else / rework anything, let me know :)

Happy0:
@mmckegg i resolved my conflicts with your changes to gathering.js yesterday and tested gatherings still render.

Let me know if I made an arse of it. Should be fine - I just copied and pasted your gathering.js file and made a render and canRender function with it again :P.

mmckegg:
@Happy0 Gonna try and get this merged in today! Will let you know.

mmckegg:
Okay, all merged! Seems to be working so far 😆

Happy0:
hurray! :D

PR #676, opened 2017-11-19
Fix typo in page heading when thread not found

jfr3000:


mmckegg:
Thanks, sorry missed this before. Will be in the next release :)

PR #679, opened 2017-11-24
Add User Font Size Setting

robjloranger:
Adds a selection for the user to change the font size from the settings
panel. Persists in localstorage and updates on changes as language and
theme. Selection range: 8,10,12,14,16,18,20 px.

Most locales updated except `ki.json` which is in english.

Font size is set in a second `style` element, appended immediately after
the theme is applied in `main-window.js`, overwriting the theme font
size. I considered finding a more clever way to replace the original
font size declaration in place but haven't got very far yet.


robjloranger:
Sure no problem, I'm actually just playing with ctrl/command -/+ with support for just = instead of having to use shift as per conversation on ssb. I'll make these changes first and then continue with shortcuts.

robjloranger:
Just a quick note, the font size is set in px on the base html/body elements in the themes. Then all text is adjusted from there. So getting the font size from the document returns a value in px. The default value is set to the current theme's base font size if no user set value is already present.

robjloranger:
Don't merge just yet, I found a bit of a bug with 3.8.0. Increasing text size breaks lots of things.

cryptix:
thanks for this contrib, rob!

On the px/pt thing: can somebody with a high res (4k/retina) screen check that this works as intended?

mmckegg:
> can somebody with a high dpi (4k/retina) screen check that this works as intended?

@cryptix yep! Works great! In modern web browsers, px aren't really pixels, they're more like points.  They're compatibility px. They'll have to come up with a new unit one day for actual screen pixels, no doubt...

mmckegg:
@robjloranger 

I think the layout issue is now [fixed](https://github.com/ssbc/patchwork/commit/689e56ed53ce16e61ad3a4145e82787c6d11960f) in latest master.

robjloranger:
works great, nice work! I think it's safe to merge. I checked from 8 - 20 px and all looks good. I like that it appears truncated but still copies the full key.

mmckegg:
Okay, merged into master! Will be in the next release :)

PR #694, opened 2017-12-15
WIP: Tray

robjloranger:
**This is not ready yet**, I need a hand with reading the settings stored with obs. I tried to replicate the dependencies from `settings.js` but was getting runtime errors.

Otherwise just needs some quick testing on windows and mac os.

### DIFF
--- a/index.js
+++ b/index.js
@@ -18,7 +18,7 @@ var windows = {
 }
 var ssbConfig = null
 var quitting = false
-
+var windowVisible = true
 // TODO: dependencies for tray settings
 
 electron.app.on('ready', () => {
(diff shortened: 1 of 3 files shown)


mmckegg:
@robjloranger 

Are you trying to read the settings from `index.js`?

That could be your problem since it is not running in the depject environment. It also expects to be a singleton, so it can't run cross process. And it depends on localStorage which isn't available in `index.js` since that is not running in a browser window. 

Unfortunately there is no easy way to handle this. 

You might have to do some kind of nasty hack where you `watch` the setting from inside the main window, and then post back to `index.js` using `electron.ipc` whenever the state changes.

See https://github.com/electron/electron/blob/master/docs/api/ipc-main.md and https://github.com/electron/electron/blob/master/docs/api/ipc-renderer.md

In `main-window.js`:

```
var yourSetting = settings.obs.get('patchwork.settingName')
watch(yourSetting, (value) => {
  electron.ipcRenderer.send('settingName-change', value)
})
```

And then in `index.js`:

```
var settingName = null
electron.ipcMain.on('settingName-change', value => {
  settingName = value
})
```

robjloranger:
Thanks @mmckegg. Not sure why I didn't get a notification for this. Anyhow I played with it a bit and it's a bit trickier than I thought. 😄 I'll have to play around with this more this weekend.

robjloranger:
I've been thinking about this and wondering about using another method of persisting user settings. Maybe writing to a json config file would be better?

mmckegg:
@robjloranger 

I think you might be right. I wonder if it should be a change to ssb-settings or if we should create something parallel? 

Would be neat to use file watchers on the json file (or at least an ipc notification on change) so that any process could update the settings and have them reflected in the other processes. This would also make multi-window patchwork much easier to do!

robjloranger:
I'm going to close this as very out of date, waiting on something better for settings management. ref: #723

PR #698, opened 2017-12-19
Added electron-reload for live reloading & dev task

prevuelta:
Optional first contribution, I added this to make dev easier, especially when updating stylesheets, let me know any thoughts.


prevuelta:
Hmmm, working for me on mac, did you run `npm run dev` instead of `npm start`?

mmckegg:
Hmm, I think this would be more useful if it remembered the page you were currently viewing. Otherwise might as well just ⌘R / Ctrl+R to reload.

I'm going to close this for now, but if we add the ability for patchwork to remember it's last page on reload, then I'd consider this.

PR #704, opened 2017-12-26
Focus existing patchwork if it is already running. Fixes #703

### LINKED ISSUE #703: Multiple instances of patchwork when double clicking the executable
When showing my dad patchwork on windows 10 earlier I noticed that if you double click the patchwork shortcut, it opens another window.

When I tried to open the appimage on my Linux laptop twice it did the same (and tried to open its own scuttlebot.)

I think this would be difficult for my dad (who is not a power user) to take into account if he had patchwork minimised :p

Maybe we can use this to stop that happening :

https://github.com/electron/electron/blob/v0.36.10/docs/api/app.md#appmakesingleinstancecallback

I might try it out if / when I have time :D

Happy0:


Happy0:
I've tested this on linux (arch linux) using `npm start` and also by building an AppImage using patchwork-builder.

I can't test it on my dad's windows machine for now as I'm not sure how to produce a windows exe :P

Happy0:
Note: it seems if you already have an _ older _ version of patchwork open, this doesn't work, so I'm not sure how the makeSingleInstance function works under the hood.

Let me know if you'd like me to try to make it work for this (more rare) use case too.

PR #709, opened 2017-12-27
Do not count your own messages as new.

### LINKED ISSUE #708: Your own replies should not count as 'new'
When you reply to someone, your reply shows up as a new post in your Public feed.

Steps to reproduce: Reply to someone, check your public feed.

This has been bothering me, so I think I'm going to try fixing this. If people agree, hopefully I'll have a pull request in a bit.

setupminimal:
This fixes #708 - your own posts no longer get the `*new` badge or count towards the number of new posts.


mmckegg:
@setupminimal 

This is probably okay to merge. The only issue will be that your posts won't appear in the public feed until someone else replies, or you manually refresh.

Not sure how I feel about that. Let's try it and see!

PR #711, opened 2017-12-27
Add an "open external link" to the context menu

reconbot:
Also cleans up a module.exports assigning that didn't work

I found that 100% of the time I would open the link in my browser to make sure it included everything relevant to the thread but I understand if we don't want to clutter the menus.

### DIFF
--- a/lib/context-menu-and-spellcheck.js
+++ b/lib/context-menu-and-spellcheck.js
@@ -62,7 +62,7 @@ function setupContextMenuAndSpellCheck (config) {
     return menu
   }
 
-  module.exports = new ContextMenuListener((info) => {
+  new ContextMenuListener((info) => {
     contextMenuBuilder.buildMenuForElement(info).then((menu) => {
       var element = document.elementFromPoint(info.x, info.y)
       while (element && !element.msg) {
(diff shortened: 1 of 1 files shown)


mmckegg:
This works, but as you mention, the menus are getting very cluttered. I think we should reorganise the menus.

For example

- Copy
  - Message ID
  - Message Text
  - External Link

Another idea is to move away from using context menus and instead add a drop down click to posts. I recall that a lot of people don't realise they can right click on the posts to get more context until someone tells them (thanks web).

I don't think I'll merge this right now, but I'll let it sit for a while as we think about how to improve the menus.

7Ds7:
I have a fork that adds the "open external link" functionality as well as a "view source" on a context menu for each post.

I do not vouch for my own code quality as i was just experimenting to wrap my head around patchwork but, feel free to have a look to see how it feels in terms of UI
https://github.com/7ds7/patchwork

https://github.com/ssbc/patchwork/compare/master...7Ds7:master

hbetts:
> Another idea is to move away from using context menus and instead add a drop down click to posts.

👍 That would definitely improve discoverability.

> I don't think I'll merge this right now, but I'll let it sit for a while as we think about how to improve the menus.

I'd love to see this merged in rather than sitting around waiting for us to find some ideal solution at a point in the distant future.

@7Ds7 does your code actually implement a visible left-clickable menu to each message? (I think that's how I'm reading it)

7Ds7:
@hbetts yes it does on the top right corner of each message, the same fork also has a profile feature to show only content created from the user without the clutter

screenshot

hbetts:
Yeah, longer-term, I think I like the more easily discoverable context menu embedded in the UI of each message. (Though would it be worth also keeping the right click context menu for those that would prefer the user experience?)

7Ds7:
@hbetts yes on that branch both functionalities are working

mmckegg:
I'd be happy to merge this if (in conjunction with #747), we removed the default external viewer out of patchwork, and instead the user had to specify one in the settings.

That way, if someone hasn't set up an external viewer, you just wouldn't show this option.

PR #720, opened 2018-01-06
style modifications to accommodate larger display sizes

rinchen:
Feel free to reject or modify this request.  :-)

I have a larger display size and the current patchwork client's limitation to 700px (less after the border intrudes) isn't very aesthetically easy to read.  These changes work fine on my system. The only odd change, mostly because I don't do this regularly, is in gathering-card.mcss. The current setup doesn't allow for auto or 100% and I discovered that with just a larger height setting, the image doesn't display completely.  I made a design choice to lock in the picture at a lower resolution (still higher than the 300px previously set).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,6847 @@
+{
+  "name": "ssb-patchwork",
+  "version": "3.8.6",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
+    "@f/zip-obj": {
+      "version": "1.1.1",
+      "resolved": "https://registry.npmjs.org/@f/zip-obj/-/zip-obj-1.1.1.tgz",
+      "integrity": "sha1-7739PokNsq2dY8dy0WImZf4GV0M="
+    },
+    "@paulcbetts/cld": {
+      "version": "github:ssbc/paulcbetts-cld-prebuilt#5bf5336c4315a24b5ae9a6019c5ae642e719f6bb",
+      "requires": {
+        "glob": "5.0.15",
+        "nan": "2.8.0",
+        "prebuild-install": "2.4.1",
+        "rimraf": "2.6.2",
+        "underscore": "1.8.3"
+      },
+      "dependencies": {
+        "glob": {
+          "version": "5.0.15",
+          "resolved": "https://registry.npmjs.org/glob/-/glob-5.0.15.tgz",
+          "integrity": "sha1-G8k2ueAvSmA/zCIuz3Yz0wuLk7E=",
+          "requires": {
+            "inflight": "1.0.6",
+            "inherits": "2.0.3",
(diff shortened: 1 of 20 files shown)


rinchen:
Full size normal image
Full size normal image

gathering-card fixed-size image
gathering-card fixed-size image

Easier to read text
Text

rinchen:
Original, unpatched for comparison

orig1
orig2
orig3

rinchen:
Another option would be for me add these as two new themes (wide light and wide dark or similar).

mmckegg:
Thanks for the PR @rinchen!

Not too keen on this being the default, but I'd be happy to have it as an option.

Rather than all the trouble of duplicating the theme, could be done as some kind of "Patch" style sheet. Add a toggle option on the Settings page that adds an additional stylesheet to the main window.

Let me know if you would like help implementing this!

rinchen:
@mmckegg 

Thanks. I could use some pointers on how to implement it. My immediate thought would be to change the hardcoded numbers in variables and allow them to be set on the settings page.

mmckegg:
@rinchen can't seem to find you in my network. What pub are you on?

---

What I would recommend is leaving the theme as is, and then creating a second override stylesheet. You could then [modify `main-window.js`](https://github.com/ssbc/patchwork/blob/master/main-window.js#L85) to load this stylesheet as well if a settings flag is in place.

There is currently this code: https://github.com/ssbc/patchwork/blob/2d8c9a07a0734a2f88fc3920b5074cb64100020a/main-window.js#L75-L81

You could modify it something like this:

```js
  var themeOverrides = require('./theme-overrides') // load stylesheets in folder and save in lookup
  var currentTheme = api.settings.obs.get('patchwork.theme', 'light')
  var currentOverrides = api.settings.obs.get('patchwork.themeOverrides', [])
  document.head.appendChild(
    h('style', {
      innerHTML: computed([currentTheme, currentOverrides], currentTheme, currentOverrides => {
        var styles = themes[currentTheme] || themes['light']
        if (Array.isArray(currentOverrides)) {
          overrides.forEach(name => {
            if (themeOverrides[name]) {
              styles += `\n${themeOverrides[name]}`
            }
          })
        }
        return styles
      })
    })
  )
```

This would allow people to create all sorts of optional style tweaks!

rinchen:
@mmckegg one.butt same as you. I sent you a few private messages today as well as friending you.

mmckegg:
I just [added a commit](https://github.com/ssbc/patchwork/commit/dde0f66d805ad433984e8d28ac8554e739f47ec7) to increase the message max-width to 800px up from 700px.

Hopefully this helps here! Any more than that and the text lines get too long for a lot of people to read. I'm going to close this PR, but feel **free to open a new one that does this full width as an option in settings**.

PR #721, opened 2018-01-07
Add French Translation

podybman:
## Linguistic problems encountered
- separation of time and the "ago" suffix makes the translation into French impossible. "2 years ago" would translate to "deux ans il y a" instead of "il y a 2 ans". So I didn't translate _ago_
## Linguistic choices
- person -> personne
- public message -> _publication_
- private message -> _message_ or _message privé_
- use of infinitive most of the time
- use of second person plural (polite form)
- channel -> _canal_
- inclusive spelling from Québec French (as it's still subject to debate in European French and simply doesn't exist in Africa)


olegskl:
Some typos:

 - [ ] "Portug**u**ais" -> "[Portugais](https://fr.wikipedia.org/wiki/Portugais)"
 - [ ] "Par défaut, Patchwork**ne** verra" -> "Par défaut, Patchwork ne verra"
 - [ ] "Rédige une réponse privée" -> "Rédige**r** une réponse privée"
 - [ ] "Rédig**u**e une réponse privée" -> "Rédig**er** une réponse privée"

olegskl:
I've never heard that a dot could be used for gender inclusivity. In fact, according to [gender neutrality in languages article](https://en.wikipedia.org/wiki/Gender_neutrality_in_languages_with_grammatical_gender#French) on Wikipedia:

> To make words or phrases gender-inclusive, French-speakers use two methods:
>
> hyphens, brackets or capital letters to insert feminine endings: étudiant-e-s, étudiant(e)s or étudiantEs; most writers avoid this practice in official titles such as Governor General and favor the next process;
>
> hendiadys containing a feminine word and a masculine word: toutes et tous, citoyennes et citoyens.

 - [ ] "s'est abonné à" -> "s'est abonné(e) à"
 - [ ] "s'est identifié.e" -> "s'est identifié(e)"
 - [ ] "s'est désabonné de " -> "s'est désabonné(e) de "
 - [ ] "L'auteur.e de ce message est peut-être en dehors de votre réseau étendu ou peut être bloqué.e." -> "L'auteur(e) de ce message est peut-être en dehors de votre réseau étendu ou peut être bloqué(e)."

podybman:
Hi @olegskl 
I'll fix the typo, I'm so sorry it went thru.

About neutrality versus inclusivity: what you're pointing at is the France official way to write. It's neutral, not inclusive. The difference is that brackets are meant to design optional and non important things (or here, people).

France is having a big debate to change it. And as a matter of fact, it's been already changed in schools (so kids are not learning that) and public administration. They are turning to the Québec spelling. See[ this document from the French administration](http://www.adelphite.fr/wp-content/uploads/2017/07/hcefh__guide_pratique_com_sans_stereo-_vf-_2015_11_05.pdf) (page 34).

So this is why I'm proposing to use the Québec writing has it has been normalized decades before what France is doing now. 

Hope this is more clear

olegskl:
Hello @podybman 

> "Grâce aux agriculteur.rice.s, aux artisan.e.s et aux commerçant.e.s, la Gaule était un pays riche."

> "Représentant.e.s des employeur.euse.s territoriaux.ales désigné.e.s et représentant.e.s..."

In my personal opinion, this is unreadable. Human beings tend to think that a dot means: "stop!". Also, it will break so many parsers that rely on dots that mark the end of sentences.

Can I at least try to persuade you to use a *middle dot* instead, please?
Like this: "s'est identifié·e". 😉

The document you link to is made by [The Haut Conseil à l'égalité entre les femmes et les hommes](http://www.haut-conseil-egalite.gouv.fr/hce/presentation-et-missions/), which is an "instance consultative", and which "formulates recommendations and opinions, and proposes reforms to the prime minister". See [Art. 9-1](https://www.legifrance.gouv.fr/affichTexteArticle.do?cidTexte=JORFTEXT000018877783&idArticle=LEGIARTI000033956508&dateTexte=&categorieLien=id) of [LOI n° 2017-86 du 27 janvier 2017 relative à l'égalité et à la citoyenneté](https://www.legifrance.gouv.fr/affichTexte.do;jsessionid=8D69073CD2AEA26C55E1AF19B5D1997D.tplgfr37s_1?cidTexte=JORFTEXT000033934948&dateTexte=20180115). They are not the authority on the French language.

The authority behind the French language (in France) is the [French Academy](https://en.wikipedia.org/wiki/Acad%C3%A9mie_fran%C3%A7aise), not the HCE you link to. The French Academy [generally opposes](http://www.academie-francaise.fr/questions-de-langue#38_strong-em-fminisation-des-noms-de-mtier-de-titres-etc-em-strong) feminisation of words in the French language, arguing that "masculin" should be considered "neutral" instead.

On 21 November 2017 the prime minister Édouard Philippe called for **abandoning inclusive writing** in all official texts. In fact, you will not find any inclusive writing in the President's [Happy New Year 2018 wishes](http://www.elysee.fr/declarations/article/v-ux-du-president-de-la-republique-pour-l-annee-201/). Just like the French Academy advises.

In my personal experience, people in France do write inclusively, mostly using parentheses (I think this is because Bescherelle and Larousse use them). I don't know anything about Québec. If Québec enforces some rules, perhaps it is wise to use fr-ca and fr-fr locales separately? With dots for Québec and middle dots or parentheses for France? What do you think?

podybman:
I'll go with the flow, don't worry.

Separating locale will be necessary but I like incremental approach. 
You are right on the fact that nothing is official yet in France. This is what I said in the first place :D

About the Académie Française, it's official but at the end, people end up writing like they were taught (which is almost never what is decided by the Académie Française). 
Let's not make it a discussion here (we should start this on SSB ;) )

I'll commit the modif to switch to something more fr-FR and we'll go from there. 

Thanks for the valuable feedback

mmckegg:
> separation of time and the "ago" suffix makes the translation into French impossible. "2 years ago" would translate to "deux ans il y a" instead of "il y a 2 ans". So I didn't translate ago

Yeah, some work needed there to make the ago strings properly localisable. That was just a quick hack! I'd happily accept a pull request to improve this.

gmarcos87:
The function that translates the time is in this line https://github.com/ssbc/patchwork/blob/da0a9b3e6b3ddd0df96ce9cedd40e2b545c2ecf4/plugs/intl/sync/i18n.js#L65.
Actually it is a workaround, because the library used for time measurements does not have its own internationalization (https://github.com/bahamas10/human).
Maybe instead of replacing the parts of the string we could translate their parts and pass them as variables to another i18n translation. And then use the singular/plural translation as past/future.

mmckegg:
@podybman 

What is the status of this PR? Do you think it is ready for including in a patchwork release?

podybman:
It's not. 
I need to rework it. Can do it tonite.
Sorry for the delay

PR #724, opened 2018-01-15
align top bar with content in fullscreen on darwin

olegskl:
## before
before-dark
before-light

## after
after-dark
after-light


mmckegg:
Looks good! And seems to work too 👍 

Merging! Will be in the next release 🎉

PR #725, opened 2018-01-15
fix typo "Seaching..." -> "Searching..."

olegskl:


mmckegg:
Thanks for the PR! Will be included in the next release!

Although, this string should probably be localised. Whoops....

PR #728, opened 2018-01-18
Optionally filter channel (un)subscribe messages from channel view

mnzaki:
Seems to be a common grievance; added a setting for it.

### DIFF
--- a/locales/en.json
+++ b/locales/en.json
@@ -197,5 +197,7 @@
 		"other": "%s people from your network replied to this message on "
 	},
 	"(you)": "(you)",
-	"Information": "Information"
+	"Information": "Information",
+	"Hide channel (un)subcribe messages": "Hide channel (un)subcribe messages",
+	"Channel Feed Options": "Channel Feed Options"
 }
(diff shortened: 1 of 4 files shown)


mmckegg:
@mnzaki 

Thanks for the pull request! Keen to merge just one thing.

I think this should also have an option for hiding the channel subscription messages from the public feed too. This case is IMO the most important.

Should be two seperate options though since you might still want to see who subscribes to a particular channel.

mnzaki:
I actually started out with one setting (that filters out subscription message from both public and channel) then changed it to channel only, because that's the obnoxious one for me: going to a channel and scrolling through a long list of 'subscribed'/'unsubscribed'. On the other hand, the subscription messages in the public feed are messages from people I'm following; it is actually interesting for me to see.

But anyway, that's a uxer preference! I added a setting for public feed filter as well.

hbetts:
👍 Looks good to me!

mmckegg:
Thanks @mnzaki! Will be in the next release 👍

PR #730, opened 2018-01-21
package.json: add implicit dependencies

mnzaki:
some deps are being require()-ed but they are not in package.json
this works because they are deps of our deps
but it is bad practice and breaks tools like https://github.com/pnpm/pnpm


hbetts:
👍 Looks good to me!

mmckegg:
Thanks @mnzaki. Merging!

PR #713, opened 2017-12-27
Show warning when the user is not following anyone, therefore may not…

Happy0:
… be able to see new content. Fixes #707.

PR #740, opened 2018-02-13
Show a warning when the user has no followers

Happy0:
A user's activity (posts, likes, etc) will not be visible to *anyone* under the following circumstances:

1. They onboard via a local network by following someone on the network. However, nobody on that network follows them back before they leave the network and open patchwork back up on another network. They can therefore see content, but nobody can see theirs.

2. They redeem a pub invite, but due to a bug (or the server spontaneously exploding at just the wrong moment, and the key is lost in the wreckage) the pub never follows the user back. This has happened to me and I have seen it happen to at least one friend (a pub invite following the pub, resulting in gossip to download new messages, but the pub not following back.)  Luckily, I was able to notice this had happened to my friend and tell them to try another pub.

I plan to investigate the scuttlebot side of this to see if I can work out why this sometimes happens, but I still think it's worth being defensive 'client side' as there may be a regression, a faulty implementation, a spontaneous server explosion, etc.

I think it's worth having a warning message for these situations. 

Note: the warning box is only displayed if the user follows someone or a pub, but is not followed back See the code comments for an explanation of why I took this approach.

Does this approach to issues 1 and 2 seem reasonable?

This pull request follows on from https://github.com/ssbc/patchwork/pull/713


locales/en.json
```diff
@@ -180,5 +180,8 @@
 	"identifies ": "identifies ",
 	" as \"": " as \"",
 	"unblocked ": "unblocked ",
-	"identified ": "identified "
+	"identified ": "identified ",
+	"Nobody will be able to see your posts until you have a follower. The easiest way to get a follower is to use a pub invite as the pub will follow you back. If you have already redeemed a pub invite and it has not followed you back, try another pub.": "Nobody will be able to see your posts until you have a follower. The easiest way to get a follower is to use a pub invite as the pub will follow you back. If you have already redeemed a pub invite and it has not followed you back, try another pub.",
```
mmckegg: Looks like there is an unused locale string here.

"Nobody will be able to see your posts until you have a follower. The easiest way to get a follower is to use a pub invite as the pub will follow you back. If you have already redeemed a pub invite and it has not followed you back, try another pub."

mmckegg: oh, you fixed this just before I commented 👍

Happy0:
Thanks @mmckegg =]. I ran 'npm test'. I like that there's a flag to make it try and fix some of the issues for you automatically :D (which I did.)

I left the unused variables reported by standard in files I didn't touch as I assume they're there for convenience.

Happy0:
Oh, I forgot to fix the formatting of 

`
var {
+  when,
+  h
+} = require('mutant')
`

I'll do that later. Goodnight! :)

Happy0:
done =]

Happy0:
Updated the pull request to check 'sync' (indexing has finished) like the recent change for the 'no followers' warning.

modules/page/html/render/public.js
```diff
@@ -267,11 +269,28 @@ exports.create = function (api) {
...
 
       return api.feed.html.followWarning(shownWhen, explanation)
     }
 
+    function noFollowersWarning () {
+      var explanation = i18n(
+        'Nobody will be able to see your posts until you have a follower. The easiest way to get a follower is to use a pub invite as the pub will follow you back. If you have already redeemed a pub invite and you see it has not followed you back on your profile, try another pub.'
+      )
+
+      // We only show this if the user has followed someone as the first warning ('you are not followed anyone')
```
hbetts: `followed` => `following`

modules/page/html/render/public.js
```diff
@@ -267,11 +269,28 @@ exports.create = function (api) {
...
       var shownWhen = computed([loading, contact.isNotFollowingAnybody],
            (isLoading, isNotFollowingAnybody) => !isLoading && isNotFollowingAnybody
-        )
+      )
 
       return api.feed.html.followWarning(shownWhen, explanation)
     }
 
+    function noFollowersWarning () {
+      var explanation = i18n(
```
hbetts: I hope at some point we can start transitioning to `const` over `var`. Though not necessarily in this change.

locales/en.json
```diff
@@ -58,128 +60,48 @@
...
 	"ago": "ago",
 	"years": "years",
 	"months": "months",
 	"weeks": "weeks",
 	"days": "days",
 	"hours": "hours",
-	"minutes": "mins",
-	"seconds": "secs",
+	"minutes": "minutes",
+	"seconds": "seconds",
```
hbetts: Are these lines related to the purpose of this change?

Happy0:
Cheers @hbetts =].

I pushed a new commit. I fixed the code comment you pointed out (well spotted.) I changed `var` to 'const' in my local functions in `modules/page/html/render/public.js` . There is a mixture of `const` and `var` in that file, so I thought it was fine to make this change in the functions I added. However, I'd rather leave changing the rests of the vars to const / let for a separate pull request unless @mmckegg wants me to change the rest of the file too.

`locales/en.json` is auto-generated, which is why you'll often see changes in it that seem unrelated with commits and pull requests @hbetts .

hbetts:
> locales/en.json is auto-generated, which is why you'll often see changes in it that seem unrelated with commits and pull requests

Ah. Thank you for the explanation. :smile: 

> However, I'd rather leave changing the rests of the vars to const / let for a separate pull request

I definitely agree with waiting until a later pull request. It's not a change directly related to the goal of this pull request.

From what else I can tell, this change looks good to me :+1:

Happy0:
\o/ :D

locales/en.json
```diff
@@ -29,13 +31,13 @@
...
-	  "other": "You share %s mutual friends with this person."
+		"one": "You share %s mutual friends with this person.",
+		"other": "You share %s mutual friends with this person."
 	},
 	"Followed by": "Followed by",
 	"You follow %s people that follow this person.": {
-	  "one": "You follow someone that follows this person.",
-	  "other": "You follow %s people that follow this person."
+		"one": "You follow %s people that follow this person.",
+		"other": "You follow %s people that follow this person."
```
mmckegg: I'm not sure why these changes were made. It is overriding the custom strings with default ones. I will attempt to roll these back before merging. Hopefully my git foo is strong enough!!

mmckegg:
Released as part of Patchwork v3.8.9 🎉

Happy0:
Yasss :D.

PR #746, opened 2018-03-02
patch-drafts for persisting drafts!

KGibb8:
Was chatting to @mixmix and he mentioned a small task porting over drafts from Patchbay into Patchwork [here](http://viewer.scuttlebot.io/%NcRUp+axz2wlm2eikbLh2gljdj44Ix0ARPB9El1oG+k=.sha256).

### DIFF
--- a/modules/message/html/compose.js
+++ b/modules/message/html/compose.js
@@ -47,7 +47,7 @@ exports.create = function (api) {
     })
 
     var draftPerstTimeout = null
-    var draftLocation = meta().location
+    var draftLocation = meta.location()
     var textArea = h('textarea', {
       'ev-input': function () {
         hasContent.set(!!textArea.value)
--- a/modules/message/html/compose.js
+++ b/modules/message/html/compose.js
@@ -53,8 +53,8 @@ exports.create = function (api) {
         hasContent.set(!!textArea.value)
         clearTimeout(draftPerstTimeout)
         draftPerstTimeout = setTimeout(() => {
-        api.drafts.sync.set(draftLocation, textArea.value)
-}, 200)
+          api.drafts.sync.set(draftLocation, textArea.value)
+        }, 200)
       },
       'ev-blur': () => {
         clearTimeout(blurTimeout)
--- a/modules/message/html/compose.js
+++ b/modules/message/html/compose.js
@@ -47,7 +47,7 @@ exports.create = function (api) {
     })
 
     var draftPerstTimeout = null
-    var draftLocation = location
+    var draftLocation = meta().location
     var textArea = h('textarea', {
       'ev-input': function () {
         hasContent.set(!!textArea.value)
--- a/modules/message/html/compose.js
+++ b/modules/message/html/compose.js
@@ -17,7 +17,10 @@ exports.needs = nest({
   'message.async.publish': 'first',
   'emoji.sync.names': 'first',
   'emoji.sync.url': 'first',
-  'intl.sync.i18n': 'first'
+  'intl.sync.i18n': 'first',
+  'drafts.sync.get': 'first',
+  'drafts.sync.set': 'first',
+  'drafts.sync.remove': 'first'
 })
 
 exports.gives = nest('message.html.compose')
@@ -43,9 +46,15 @@ exports.create = function (api) {
       }
     })
 
+    var draftPerstTimeout = null
+    var draftLocation = location
     var textArea = h('textarea', {
       'ev-input': function () {
         hasContent.set(!!textArea.value)
+        clearTimeout(draftPerstTimeout)
+        draftPerstTimeout = setTimeout(() => {
+        api.drafts.sync.set(draftLocation, textArea.value)
+}, 200)
       },
       'ev-blur': () => {
         clearTimeout(blurTimeout)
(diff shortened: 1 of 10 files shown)


cryptix:
Oh great!

This reminds me to give the ssbc crew access to the npm package. Is there an easy way to do that?

KGibb8:
Whoops! Nice spot @cryptix, have changed :smile:

KGibb8:
What's the testing convention, if there is one, in Patchwork?

mixmix:
This is rad, thanks @KGibb8 ! I only skimmed it but looks like you've got the idea. @mmckegg do you want to review this or shall I handle it and merge it if it's good.

The drafts plugin is something I want to revisit and make more fully featured at some stage.
e.g. add a page which shows all of the drafts you had in progress, and to jump to them.

mmckegg:
I’d like to review it :)

Will try and check it out tomorrow sometime.

modules/message/html/compose.js
```diff
@@ -43,9 +46,15 @@ exports.create = function (api) {
       }
     })
 
+    var draftPerstTimeout = null
+    var draftLocation = location
```
mmckegg: In patchwork, `location` is always the same (doesn't change on navigate) so the same draft appears in every box right now. I'm guessing this isn't intentional.

You should probably use something from `meta`. Either pass a key in from the `public`, `message`, and `channel` views (`location: id`), or use an existing key.

mixmix:
In the future I think drafts should be part of an extracted composr

modules/message/html/compose.js
KGibb8: @mmckegg thanks I'm still figuring out the overall structure. Will have a look tomorrow.

KGibb8: Changed to use `${path}/${id}` to ensure uniqueness. Doesn't use the root message id so saves different drafts for different replies in threads. 

meta now always returns a `Struct`. [As its already used once](https://github.com/ssbc/patchwork/blob/master/modules/page/html/render/message.js#L32), it makes sense to always use rather than check to see whether or not to call the function.

KGibb8:
@mixmix perhaps extracting this into a composer is a little task I can do to improve my knowledge. Maybe you can advise?

modules/page/html/render/channel.js
```diff
@@ -36,7 +36,11 @@ exports.create = function (api) {
         ])
       ]),
       api.message.html.compose({
-        meta: {type: 'post', channel},
+        meta: Struct({
```
mmckegg: You can avoid turning these into structs by using `mutant/resolve`:

```js
var resolve = require('mutant/resolve')
var draftLocation = resolve(meta.location) 
// now it doesn't matter if meta.location is an observable or just an ordinary key.
```

modules/page/html/render/private.js
```diff
@@ -19,7 +20,10 @@ exports.create = function (api) {
     const i18n = api.intl.sync.i18n
     var id = api.keys.sync.id()
     var compose = api.message.html.compose({
-      meta: {type: 'post'},
+      meta: Struct({
+        type: 'post',
+        location: `${path}/${id}`
```
mmckegg: The `id` in this case is actually the public key of the current user.

It potentially could be useful to include this, that way if later we add the ability to switch identities, drafts would be stored per identity. So you probably want to add the user's public key to all draft locations, or none, instead of just some.

modules/page/html/render/channel.js
KGibb8: Nice thanks.

modules/page/html/render/private.js
KGibb8: Seems to make sense to include it, I added it to standardize how it looks across the different views . So mainly for the sake of clarity in the code. However, you're right, it currently makes no difference if this is included. Good to see a future use case.

KGibb8:
After exploring a bit more, seems like using the entire meta obj seemed excessive. Harder to recall information...

mixmix:
hey keen for this to keep a good feel of momentum - @KGibb8 hit me up if you need any help

KGibb8:
Thanks @mixmix this hasn't been my focus last couple of days. Lets chat soon.

modules/page/html/render/private.js
```diff
@@ -19,7 +19,10 @@ exports.create = function (api) {
     const i18n = api.intl.sync.i18n
     var id = api.keys.sync.id()
     var compose = api.message.html.compose({
-      meta: {type: 'post'},
+      meta: {
+        type: 'post',
+        location: `${path}/${id}`
```
mmckegg: @KGibb8 

Hmm, probably the `location` shouldn't be inside of meta. Should probably just be in the root options passed to `compose`. Otherwise it will be included in the final published message.

This is getting real close!

KGibb8: Whoops! Ofc! :man_facepalming:

KGibb8:
Going to build a page for drafts and open different PR.

mixmix:
@matt you out a bit of work in replying, was this anywhere near being
merged?

@kieran can you leave a note for others who contributed time to this why
you're closing it

mmckegg:
@mixmix I would have been happy to merge at https://github.com/ssbc/patchwork/pull/746/commits/d34ad5a41ad19886bfe4dd7cde719e0bb8f97b60

But it has since diverged from there. I think @KGibb8 was exploring putting together a drafts page, although not sure why they closed it. 

(oh and @mixmix I think you @-mentioned the wrong person -- sorry other person!)

mixmix:
sorry about that - i was responding via email. I'm too used to subjective at-mentions in scuttlebutt. All other at-mentions suck now

KGibb8:
Hey happy to reopen this at a point where it made sense. I carried on hacking at it to learn more and pushed stuff up here when it wasn't really relevant to this PR.

PR #747, opened 2018-03-06
[wip] Add the option to change the url of the external viewer.

gmarcos87:
I think it is convenient to be able to configure this option easily (currently it can be done by editing .ssb/config).
What I did was add an option to the configuration panel and use that url when right clicking on a post.

captura de pantalla de 2018-03-06 10-49-40

### DIFF
--- a/lib/context-menu-and-spellcheck.js
+++ b/lib/context-menu-and-spellcheck.js
@@ -5,7 +5,7 @@ var ref = require('ssb-ref')
 
 module.exports = setupContextMenuAndSpellCheck
 
-function setupContextMenuAndSpellCheck (config) {
+function setupContextMenuAndSpellCheck (config, settings) {
   window.spellCheckHandler = new SpellCheckHandler()
   window.spellCheckHandler.attachToInput()
 
@@ -68,6 +68,7 @@ function setupContextMenuAndSpellCheck (config) {
       while (element && !element.msg) {
         element = element.parentNode
       }
+      var viewerUrl = settings.get('patchwork.viewerUrl','https://viewer.scuttlebot.io')
 
       menu.append(new MenuItem({
         label: 'Inspect Server Process',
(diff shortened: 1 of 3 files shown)


cryptix:
I welcome this change very much!

I'm not sure how to feel about the unrelated indenting in the locale file. I prefer seperate PRs but I think it's fine in this case?

gmarcos87:
@cryptix Sorry, that was my editor and I didn't realize it. I already removed it.
About the default url my question was where to put it. Perhaps not on either side, but in a third place to guarantee its execution at least once. Where do you think it's more accurate?

mmckegg:
I'm starting to think that Patchwork should probably not have an external viewer by default.

What if there was an option called "Specify External Viewer" that when ticked made the box appear to enter the pub url?

If someone hasn't specified the viewer, the "Copy External Link" option would not show.

hbetts:
> I'm starting to think that Patchwork should probably not have an external viewer by default.

@mmckegg what problem would not having it enabled by default solve? (Trying to know why I wouldn't always just want something to be available)

hbetts:
> Having a hardcoded external pub in patchwork violates the decentralised nature of ssb.

Yep, that answers my question @mmckegg. Thank you!

I've definitely felt references to fixed, external, URLs are an antipattern in SSB. 

> The external link will only work if you are friends with someone who is on that hardcoded pub.

Then there's that significant limitation. I wonder if we introduced a lot of complexity, if we could make it where if you spun up an `ssb-viewer` using its own account, but linked it to your primary account with `sameAs`, that your client would see you have an account with _viewer_ capabilities, so it would enable the _External Viewer_ option in your client. Basically, if you want to link to content you have access to, you must host your own viewer. That also seems to make sense because, as you mentioned, linking to content only really works if that content is accessible to the `ssb-viewer`.

Anyways, that's way out of scope, but it's what comes to mind when I think about the relationship between the decentralized world of ssb and the fixed world of WWW.

So back on topic, a checkbox seems great!

PR #748, opened 2018-03-15
[wip] Integrating Tags

wittjosiah:
The goal of this PR is to get Patchwork setup to be able to view and edit public tags using [ssb-tags](https://github.com/wittjosiah/ssb-tags) and [patch-tag](https://github.com/wittjosiah/patch-tag).

TODO:
- [x] Potentially pull more modules out into patch-tag
- [x] Review and publish both patch-tag and ssb-tags
- [x] i18n
- [x] Dark Theme Styles


modules/page/html/render/tag.js
```diff
@@ -0,0 +1,82 @@
...
+      h('div.side', [
+        h('h2', 'Your Tags'),
+        map(
+          tags,
+          tag => computed(
+            tag,
+            ({ tagId, tagName }) =>
+              h('a', {
+                'ev-click': () => selectedTagId.set(tagId),
+                style: { 'margin-bottom': '.3rem', display: 'inline-block' }
```
mixmix: This should be done with mccs instead of inline styles

modules/page/html/render/tag.js
```diff
@@ -0,0 +1,82 @@
...
+        : []
+    )
+
+    var prepend = [
+      h('PageHeading', [
+        h('h1', [ h('strong', 'Tags'), ' from you' ]),
+      ])
+    ]
+
+    return h('div.SplitView', [
```
mixmix: What you've written is correct, but more normal mutant style would be to write `h('SplitView', [`

modules/page/html/render/tag.js
```diff
@@ -0,0 +1,82 @@
...
+        ? map(
+            api.tag.obs.taggedMessages(myId, tagId),
+            msgId => api.message.obs.get(msgId)
+          )
+        : []
+    )
+
+    var prepend = [
+      h('PageHeading', [
+        h('h1', [ h('strong', 'Tags'), ' from you' ]),
```
mixmix: this isn't going to work for i18n unfortunately ): as you can't gaurentee noun position

modules/page/html/render/tag.js
```diff
@@ -0,0 +1,82 @@
...
+    return h('div.SplitView', [
+      h('div.side', [
+        h('h2', 'Your Tags'),
+        map(
+          tags,
+          tag => computed(
+            tag,
+            ({ tagId, tagName }) =>
+              h('a', {
+                'ev-click': () => selectedTagId.set(tagId),
```
mixmix: :warning: This means the page won't refresh, but it also means you're side-stepping the history, so that when people click back they won't get the last thing they were on.

I think you should push into the history (or however it's done here in patchwork). If you have problems with the view reloading being a jolt then you might need to build some caches, but figure that out later

mixmix: hmmm that's gonna change (and simplify) a bunch of your logic removing that observeable

modules/sheet/editTags.js
```diff
@@ -0,0 +1,183 @@
...
+      input.addEventListener('suggestselect', onSuggestSelect)
+
+      return {
+        content: [
+          stagedTags,
+          h('EditTags', input)
+        ],
+        footer: [
+          h('button.save', {
+              'ev-click': () => {
```
mixmix: this could just be `'ev-click': publish` (unless the order of onSave / close matters)

modules/sheet/editTags.js
```diff
@@ -0,0 +1,183 @@
...
+      'apply': 'first',
+      'create': 'first',
+      'name': 'first'
+    },
+    'html': {
+      'tag': 'first'
+    },
+    'obs': {
+      'messageTagsFrom': 'first',
+      'allTags': 'first'
```
mixmix: Matt might have a different preference but I prefer all depject things to be flat to make searching the project for uses easier. 
i.e. `tag.obs.messageTagsFrom` instead of that nested thing.

I haven't looked at the rest of the project recently to gauge style to be fair

modules/sheet/tags.js
```diff
@@ -0,0 +1,217 @@
...
+          type: 'search',
+          placeholder: 'filter tags',
+          'ev-input': function (ev) {
+            currentFilter.set(ev.target.value)
+          },
+          hooks: [FocusHook()],
+          style: {
+            'float': 'right',
+            'font-size': '100%'
+          }
```
mixmix: this inline styling needs to come out.
Let me know if you'd like an intro to mcss at all

modules/sheet/tags.js
```diff
@@ -0,0 +1,217 @@
...
+      }
+    })
+
+    return content
+  }
+
+  function renderTagBlock (tags) {
+    var yourId = api.keys.sync.id()
+    return [
+      h('div', {
```
mixmix: this should be `h('TagList',` and the function should be called `TagList` IMO.

I think when you start extracting MCSS you'll want to name some other components too. You might even want to extract the components into different files to make this file a little more trim

modules/sheet/tags.js
```diff
@@ -0,0 +1,217 @@
...
+      ]),
+      renderTaggersBlock(filteredIds)
+    ])
+  }
+
+  function renderTaggersBlock (profiles) {
+    var yourId = api.keys.sync.id()
+    profiles = api.profile.obs.rank(profiles)
+    return [
+      h('div', {
```
mixmix: `h('ProfileList',`

ProfileList might be a bit generic for mcss purposes (you don't want to collide with others). Maybe mention tags in the classname somewhere?

modules/tag/html/tag.js
```diff
@@ -0,0 +1,57 @@
...
+          'background-color': backgroundColor,
+          'color': fontColor
+        }
+      },
+      [
+        h('span', tagName),
+        removeTag
+      ]
+    )
+  }})
```
mixmix: :question: - your code looks good overall, I find your code quite spread out which makes it hard to read much without scrolling. e.g. the above is using pretty much as many lines as it could. Here's the same thing in significantly less lines and still quite readable IMO

```js
const style = {
  'background-color': backgroundColor,
  'color': fontColor
}
return h('Tag', { style }, [
  h('span', tagName),
  removeTag
])

```

mixmix:
Looking really good @wittjosiah
I think I'd love to recommend separating view and data logic more - I'm currently leaning strongly in favour of having outside modules like scuttle-tags prepare data structures, and handle publishing etc.
Not sure if that's needed for this release, it's really up to Matt at the moment. I think less logic in a PR does make it easier to merge though - it means that if you need to fix anything you can go off and fix a module and then the PR here might just be a version bump.
 
@mmckegg I've given this a first pass review. I don't know if that's useful - we haven't had a maintenance conversation. I notice that makes it a little awkward for me to support you and that my review here may not be useful or even welcome! I'd love to talk about maintainership of Patchwork in a proactive way so that you don't have to be the bottleneck. There's a much deeper conversation about how we do code in such a way that we can easily say yes.... like plugins.

mixmix:
oh yeah - I would recommend `scuttle-tag` over `patch-tag` - difference being that scuttle-tag is a module that could be used by patchfoo. This is a bigger conversation because it would necessitate extracting some of the socialValue logic (from patchcore) ... but I think that would be a great module in itself too... (we've got a line of yaks here, we're gonna need a bunch of help shaving them!)

modules/page/html/render/tag.js
wittjosiah: This wasn't actually even being used, so I just removed it

modules/sheet/tags.js
wittjosiah: I had actually used `ProfileList` so that it would pick up already existing styling. I copied the styling for now though, its duplicated but probably less likely to get accidentally broken if changes are made to one list. Maybe could be refactored out of both sheets at some point.

wittjosiah:
@mixmix Thanks for taking the time to review! I think I addressed most of your comments with some changes :)

mmckegg:
Hey @wittjosiah, thanks for getting started on this! Although, I have a few recommendations to help this get merged. 

I've never documented any PR guidelines for Patchwork anywhere, but since your PR is quite a large one, I thought [now might be a good time](https://github.com/ssbc/patchwork/blob/master/contributing.md). Also, sorry if you feeling like you are getting conflicting information. We are still figuring this stuff out, and the opinions regarding best practice have probably diverged quite a bit.

> # Patchwork Contribution Guidelines
>
> ## Please ensure your PR conforms to [JavaScript Standard Style](https://standardjs.com/)
>
> You can run `npm test` or `npx standard` to run the linter. A lot of small issues can be automatically fixed using `npx standard --fix`.
> 
> ## Feel free to use [depject](https://github.com/depject/depject) inside of the Patchwork repo, but please don't merge in depject modules from other packages. 
> 
> [patchcore](https://github.com/ssbc/patchcore) and [patch-settings](https://github.com/mixmix/patch-settings) are the only external packages that [merge in depject modules](https://github.com/ssbc/patchwork/blob/master/main-window.js#L23) (and that is because of legacy reasons). Please don't treat this as best practice. 
> 
> I recommend doing as much as possible in the Patchwork repo rather than splitting into a bunch of modules. This makes it much easier to maintain and update the overall application. Only split things into modules that are clearly reusable components (algorithms, widgets, indexes) and use `require` to include them.
> 
> _Originally depject was included in Patchwork to ease plugin creation, but I have realised since that this was not the best approach. When used across packages, it tends to create confusing hard to maintain APIs and brittle namespacing._
> 
> **Please consider depject deprecated in Patchwork!** It will be replaced with something else in the near future, and I want to keep as much of it contained in the main patchwork repo to assist with migration. 

Pretty sure I will think of more things to include, but that's all I have for now. 

I'm keen to get tags merged into Patchwork, so please let me know if you need any assistance with restructuring to address the depject stuff.

modules/sheet/display.js
```diff
@@ -14,6 +14,8 @@ exports.create = function () {
 
     document.body.appendChild(container)
 
+    if (mounted) mounted()
```
mixmix: call `afterMount` or `onMount`

modules/sheet/editTags.js
```diff
@@ -0,0 +1,175 @@
...
+      )
+
+      const messageTagsView = map(
+        filteredMessages,
+        tag => computed(tag, t => api.tag.html.tag(t, () => tagsToRemove.push(t.tagId)))
+      )
+      const tagsToApplyView = map(
+        tagsToApply,
+        tag => api.tag.html.tag(tag, () => tagsToApply.delete(tag))
+      )
```
mixmix:
```js
const Tag = api.tag.html.tag
```
```js
const tagsToApplyView = map(tagsToApply, Tag(tag, () => tagsToApply.delete(tag)))
```

modules/sheet/tags/renderTaggers.js
```diff
@@ -0,0 +1,92 @@
...
+            title: id
+          }, [
+            h('div.avatar', [api.about.html.image(id)]),
+            h('div.main', [
+              h('div.name', [ api.about.obs.name(id) ])
+            ]),
+            h('div.buttons', [
+              api.contact.html.followToggle(id, {block: false})
+            ])
+          ])
```
mixmix:
```js
+          return h('a.profile', 
+            { href: id, title: id }, 
+            [
+              h('div.avatar', [api.about.html.image(id)]),
+              h('div.main', [
+                h('div.name', [ api.about.obs.name(id) ])
+              ]),
+              h('div.buttons', [
+                api.contact.html.followToggle(id, {block: false})
+              ])
+          ])


```

mixmix:
This is looking _really good_. Tested and it's working (with npm modules linked in).
Recommend a couple of bug fixes then get Matt's eyes on this for a review

Serkan-devel:
Doesn't a dark theme already exist?

wittjosiah:
@Serkan-devel yes, that TODO is in reference to making dark theme styles specifically for the new tag UI elements

main-window.js
```diff
@@ -114,6 +114,7 @@ module.exports = function (config) {
         dropTab(i18n('More'), [
           getSubscribedChannelMenu,
           [i18n('Gatherings'), '/gatherings'],
+          ['Tags', '/tags'],
```
mixmix: i18n

modules/message/html/tag.js
```diff
@@ -0,0 +1,17 @@
...
+})
+
+exports.gives = nest('message.html.action')
+
+exports.create = (api) => {
+  return nest('message.html.action', msg => {
+    return h('a.tag', {
+      href: '#',
+      'ev-click': () => api.sheet.editTags({ msgId: msg.key }, console.log)
+    }, 'Tag')
```
mixmix: i18n

modules/message/html/tag.js
```diff
@@ -0,0 +1,17 @@
+var { h } = require('mutant')
```
mixmix: This file is in the wrong folder.
Should be in modules/message/hmtl/action/tag.js

modules/page/html/render/tag.js
```diff
@@ -0,0 +1,57 @@
...
+    const urlId = path.substr(6)
+    const currentTagId = urlId && decodeURIComponent(urlId)
+    const myId = api.keys.sync.id()
+    const ScuttleTag = TagHelper(api.sbot.obs.connection)
+
+    const tags = map(ScuttleTag.obs.allTagsFrom(myId), tagId => ({
+      tagId,
+      tagName: api.about.obs.name(tagId)
+    }))
+    const name = currentTagId ? api.about.obs.name(currentTagId) : 'Select A Tag'
```
mixmix: i18n

modules/page/html/render/tag.js
```diff
@@ -0,0 +1,57 @@
...
+    const tagMessages =
+      currentTagId ? map(ScuttleTag.obs.messagesTaggedByWith(myId, currentTagId), msgId =>
```
mixmix: _personal preference_ :open_hands: 

I think I'd prefer this api to be `ScuttleTag.obs.taggedMessages({ tagger:myId, tagId: currentTagId  })`

then if you omit `tagger` or `tagId` ScuttleTag handles what to do (which this current pattern suggests it might already be doing?). It could also handle the getting of the messages on its side to save some mess this side

modules/page/html/render/tag.js
```diff
@@ -0,0 +1,57 @@
...
+              { href: `/tags/${encodeURIComponent(tagId)}`, title: tagId },
+              api.tag.html.tag({ tagName, tagId }, null))))
+      ]),
+      h('div.main', [
+        h('Scroller', [
+          h('h2', name),
+          h('section.messages', [
+            map(tagMessages, msg =>
+              computed(msg, msg => {
+                if (msg && !msg.value.missing) {
```
mixmix: ideally tagMessages (and ScuttleTag upstream) could handle this filtering detail.
This logic is probably going to be duplicated across clients otherwise

modules/tag/obs/suggest.js
```diff
@@ -0,0 +1,79 @@
...
+      filtered.push({
+        title: 'Click or press , to create a new tag',
+        value: word,
+        tagId: false
+      })
+      return filtered
+    }
+  })
+
+  function loadSuggestions (stagedTagIds) {
```
mixmix: stagedTagIds not used in this fn

modules/tag/obs/suggest.js
```diff
@@ -0,0 +1,79 @@
...
+  var matches = api.intl.sync.startsWith
+
+  return nest('tag.obs.suggest', function (stagedTagIds) {
+    loadSuggestions(stagedTagIds)
+    return function (word) {
+      var filtered
+      if (!word) {
+        filtered = suggestions().slice(0, 200)
+      } else {
+        filtered = suggestions().filter((item) => {
```
mixmix: this method shouldn't be in tag.obs.suggest because it's not returning and obs, it's just returning a standard object as far as I can tell.

I would recommend making this method in `tag.obs.suggest`
I also think you should extract most of this out into ScuttleTag, and just have a skeleton left here so you can run loadSuggestions ad have it populating this `suggestions` cache. For myself I just make sure somewhere in the app I call this function to kick off the suggestions population.

modules/sheet/editTags.js
```diff
@@ -0,0 +1,137 @@
...
+      }
+
+      function onMount () {
+        input.focus()
+        const appliedTagIds = map(filteredMessages, tag => tag.tagId)
+        const applyTagIds = map(tagsToApply, tag => tag.tagId)
+        const stagedTagIds = computed([ appliedTagIds, applyTagIds ], (a, b) => concat(a, b))
+        const getTagSuggestions = api.tag.obs.suggest(stagedTagIds)
+        addSuggest(input, (inputText, cb) => {
+          cb(null, getTagSuggestions(inputText))
```
mixmix: I've changed all my suggestions methods to be async, like so this line for me is : 

```
getTagSuggestions(inputText, cb)
```

modules/sheet/editTags.js
```diff
@@ -0,0 +1,137 @@
...
+        }
+      }
+
+      function onSave () {
+        // tagsToCreate
+        forEach(tagsToCreate(), tag => {
+          ScuttleTag.async.create(null, (err, msg) => {
+            if (err) return
+            ScuttleTag.async.name({ tag: msg.key, name: tag }, cb)
+            ScuttleTag.async.apply({ tagged: true, message: msgId, tag: msg.key }, cb)
```
mixmix: I'm not sure, but you perhaps don't want to call cb twice here?
Don't know what tend to be passed in here.
You could run these in parallel then run the cb once after they both finish with something like https://github.com/feross/run-parallel ... or just run them one after the other like nest callbacks

modules/sheet/editTags.js
```diff
@@ -0,0 +1,137 @@
...
+          })
+        })
+        // tagsToApply
+        forEach(tagsToApply(),
+          tag => ScuttleTag.async.apply({ tagged: true, message: msgId, tag: tag.tagId }, cb))
+        // tagsToRemove
+        forEach(tagsToRemove(),
+          tagId => ScuttleTag.async.apply({ tagged: false, message: msgId, tag: tagId }, cb))
```
mixmix: right, that cb is getting hammered!

modules/sheet/tags/renderTags.js
```diff
@@ -0,0 +1,71 @@
...
+exports.gives = nest('sheet.tags.renderTags')
+
+exports.create = function (api) {
+  const i18n = api.intl.sync.i18n
+  return nest('sheet.tags.renderTags', function (ids, select) {
+    var currentFilter = Value()
+    var tagLookup = mutantLookup(ids, (id) => {
+      return [id, api.about.obs.name(id)]
+    })
+    var filteredIds = computed([ids, tagLookup, currentFilter], (ids, lookup, filter) => {
```
mixmix: I notice this code has been repeated. You could extract it to scuttleTag?

modules/sheet/editTags.js
wittjosiah: @mixmix Yeah, I remember writing this and thinking that this callback was gonna get hit a ton. At the time I just wanted to get it working so it wasn't a huge issue.

Taking a look at it now, setting it up to that these all run and callback once should definitely be possible but could be a bit complex I think. `tagsToApply` and `tagsToRemove` aren't that bad, but `tagsToCreate` needs to first create all the tags, then take those tags and name/apply them.

I'm not used to coding with callbacks so while I'm sure its possible how to set it up is not immediately obvious to me. I'll revisit it soon hopefully, but if you have any more pointers that would be helpful as well.

modules/sheet/tags/renderTags.js
wittjosiah: Which code specifically are you referring to here? Filtering the tags?

mmckegg:
@wittjosiah I still need to review the tagging dialog and the feed integration.

modules/page/html/render/tag.js
```diff
@@ -0,0 +1,110 @@
...
+    return h('SplitView -tags', [
+      h('div.side', [
+        h('h2', i18n('Tags Used By You')),
+        map(tagsUsedByYou, tag => computed(tag, HtmlTag)),
+        h('div.heading', [
+          h('h2', i18n('Community Tags')),
+          h('span.tagInfo', when(showMostActive, i18n('(most active)'), i18n('(recent)')))
+        ]),
+        h('div.tagControls', [
+          h('button.mostActive', {
```
mmckegg: These toggle buttons (and the ones below) should probably follow the same styles as scuttle-poll:

https://github.com/ssbc/patchwork/pull/814/files#diff-a1fa57760d9f0bdb1412ec84fef581aeR70

wittjosiah:
@mmckegg I think that addresses all the stuff you mentioned. Let me know if I missed anything!

mmckegg:
@wittjosiah haven't had a chance to try yet, but the code looks good!

modules/sheet/editTags.js
wittjosiah: Finally got around to looking at this again, and seems to be working great with run-parallel, thanks for the pointer @mixmix 😄

mixmix:
Looking at this branch and thinking : **I want to merge this by next Friday**

This has dragged on quite a while and I feel like we're blocking and disrespecting Josiah's work now. I'm pretty tempted to push merge and just work out the details once it's in master :japanese_goblin:

mixmix:
Have organised an in person pairing session with Matt on this coming weds.
Will aim to merge this

wittjosiah:
Sounds good @mixmix. I've added dark theme styles and updated from master to remove conflicts which have arisen.

styles/light/confirmSuggest.mcss
```diff
@@ -0,0 +1,28 @@
...
+      }
+      :last-child {
+        border: 0
+      }
+      img {
+        height: 20px
+      }
+    }
+  }
+}
```
christianbundy: I think this may have been accidentally, do you think this can/should  be deleted? The same file was included with the "standard" hyphenated filename in https://github.com/ssbc/patchwork/commit/9f9a4c5c0cb02112b8ed10ddfcd2ebe06c052ef5, so I think `confirmSuggest.mcss` can be safely deleted.

styles/light/editTags.mcss
```diff
@@ -0,0 +1,13 @@
...
+}
+
+EditTags {
+  margin: 1rem 1rem 1rem 0
+  display: inline-block
+  input.tags {
+    font-size: 1rem
+    padding: .4rem
+  }
+}
```
christianbundy: Same as https://github.com/ssbc/patchwork/pull/748#discussion_r216378274, this also looks like a duplicate file.

styles/light/confirmSuggest.mcss
wittjosiah: @christianbundy nice catch! I thought I deleted this and `editTags.mcss` I will remove them.

christianbundy: No problem! I was curious what the diff from your branch would look like for #850, but I got a few "file not found" errors and noticed the change. The results are [here](https://github.com/ssbc/patchwork/commit/50b43020b1d5b043166adff255149c690420789e), if you want to double-check your dark theme work. Looks great!

modules/sheet/tags/renderTags.js
christianbundy: I think he may be referring to `filteredIds`, repeated [here](https://github.com/ssbc/patchwork/pull/748/files#diff-e077b5d0ca4384194ede4fe34c2c19beR26).

mmckegg:
Note to @mixmix and Matt (when merging): Add a `normalize` to tag to strip off weird characters including `#` at the start of a tag.

mixmix:
It's merged Josiah. Great work, this is gonna be a game changer

PR #754, opened 2018-03-27
Tweak german translation

optikfluffel:


cryptix:
Vielen dank! All of this LGTM.

PR #755, opened 2018-03-27
Patchdrafts (reopened...)

KGibb8:
See https://github.com/ssbc/patchwork/pull/746

### DIFF
--- a/main-window.js
+++ b/main-window.js
@@ -20,6 +20,7 @@ module.exports = function (config) {
     require('./modules'),
     require('./plugs'),
     require('patch-settings'),
+    require('patch-drafts'),
     require('patchcore'),
     require('./overrides')
   )
--- a/modules/message/html/compose.js
+++ b/modules/message/html/compose.js
@@ -47,7 +47,7 @@ exports.create = function (api) {
     })
 
     var draftPerstTimeout = null
-    var draftLocation = meta().location
+    var draftLocation = meta.location()
     var textArea = h('textarea', {
       'ev-input': function () {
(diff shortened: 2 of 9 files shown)


mmckegg:
Hey @KGibb8,

I'm keen to merge this. The only hold up is that this depends on `patch-drafts` which depends on [merging in depject from an external package](https://github.com/KGibb8/patchwork/blob/67322448987472eaa84be951dd1e6aa1e2e9b2e2/main-window.js#L23).

I recently wrote a post about deprecating external depject packages here: [%+c4BZ9I0XGYFV6AjmHCEa3+rn6PYa9QWm365GEie5TE=.sha256](https://viewer.scuttlebot.io/%25%2Bc4BZ9I0XGYFV6AjmHCEa3%2Brn6PYa9QWm365GEie5TE%3D.sha256).

> Going forward, I would like to avoid merging in external depject modules into patchwork. Instead, any external deps should be done using standard require (but generally avoided unless the code is a reusable algorithm or widget). This will make it a lot easier to replace depject (or at least refactor the module structure considerably) with something more suited to plugins at a later stage, and my life a lot simpler in the meantime when merging PRs and refactoring stuff (e.g. adding sameAs support).

The simple answer to solving this is to implement the draft persistence directly in this PR. Or maybe patch-drafts could be updated to _not be depject based_, or provide an alternate api? Any thoughts @cryptix?

cryptix:
> Any thoughts @cryptix?

to be honest: most of the patch-drafts package is boilerplate around the localStorage key-value store.

PR #756, opened 2018-03-29
npm test up to date, standard js conformant

Space-Ed:
npm test wouldn't run, missing package,

added the package and ran standard, it gave some warns so I fixed them

the app still runs but see if anything changed was needed


mmckegg:
Thanks @Space-Ed, merged!

PR #757, opened 2018-03-29
Added linear-gradient mask to channel links in active channels list

mroseman95:
Same lineary-gradient style used in Who to Follow list.

This fixes a bug where channel names would overflow outside their box if font size was to
large.

Before: 
channel_name_overflow

After:
channel_name_overflow_fixed

Signed-off-by: Matthew Roseman <mroseman95@gmail.com>


styles/light/all-channels.mcss
```diff
@@ -48,6 +48,7 @@ AllChannels {
       flex: 1
       white-space: nowrap;
       min-width: 0;
+      -webkit-mask-image: linear-gradient(90deg, rgba(0,0,0,1) 90%, rgba(0,0,0,0));
```
mmckegg: And here!

styles/dark/all-channels.mcss
```diff
@@ -60,6 +60,7 @@ AllChannels {
       flex: 1
       white-space: nowrap;
       min-width: 0;
+      -webkit-mask-image: linear-gradient(90deg, rgba(0,0,0,1) 90%, rgba(0,0,0,0));
```
mmckegg: This is the source of the scroll perf hit. Seems to be okay on the Public sidebar, but when rendering thousands of them, things are getting pretty janky on my machine.

screen shot 2018-04-03 at 12 42 02 pm

I suggest replacing this code with a `text-overflow: ellipse` or something similar. For example:

```css
      text-overflow: ellipsis;
      overflow: hidden;
```

mroseman95:
I couldn't find a way to only apply the linear gradient to elements with overflowed text.
I did see some [documentation](https://developer.mozilla.org/en-US/docs/Web/CSS/text-overflow#fade) for a `text-overflow: fade`, but it isn't supported.

So I implemented the `text-overflow: ellipsis`.

I also changed the `overflow-y: auto` to `overflow: auto` to remove the unnecessary horizontal scrolling I was getting in the AllChannels page.

mroseman95:
The only way I can think of to only apply the gradient to overflowed text would be to somehow check for overflow in js, and add a class to apply the gradient.

mmckegg:
Thanks @mroseman95, merged!

`text-overflow: fade` would be perfect. For now let's just wait and see if it gets supported. It is fairly rare for channels to be that long unless font size is set a lot higher than the default.

PR #763, opened 2018-04-06
add rendering for blog messages

mixmix:
**PR-type**: feature (minor version change?)

**what it does**: adds a plug which renders "blog" type messages. By default, the blog is loaded in a 'collapsed' form as a BlogCard. This means the blog doesn't take up as much space, and also gives time for the blob (where the blog content) to be fetched gracefully underneath.
When you click the card it expands the blog into it's full form

## Blog card  (collapsed form)

selection_568

:heavy_check_mark:  There's styling for light and dark BlogCards

Here's what it looks like with replies:

selection_570

## Blog (expanded form)

selection_569

### DIFF
--- a/modules/feed/html/rollup.js
+++ b/modules/feed/html/rollup.js
@@ -244,6 +244,7 @@ exports.create = function (api) {
       var renderedMessage = api.message.html.render(item, {
         compact: compactFilter(item),
         includeForks: false, // this is a root message, so forks are already displayed as replies
+        renderAsCard: true,
         priority: getPriority(item)
       })
 
--- a/modules/feed/html/rollup.js
+++ b/modules/feed/html/rollup.js
@@ -244,7 +244,7 @@ exports.create = function (api) {
       var renderedMessage = api.message.html.render(item, {
         compact: compactFilter(item),
         includeForks: false, // this is a root message, so forks are already displayed as replies
-        renderAsCard: true,
+        inRollup: true,
         priority: getPriority(item)
       })
 
--- a/package.json
+++ b/package.json
@@ -56,6 +56,7 @@
     "pull-ping": "^2.0.2",
     "pull-pushable": "^2.0.1",
     "pull-stream": "~3.6.0",
+    "scuttle-blog": "^1.0.0",
     "scuttlebot": "github:ssbc/scuttlebot#22d894d03235a96f1cad8e836b55ca57eb25d17f",
     "secret-stack": "4.0.1",
     "sorted-array-functions": "~1.0.0",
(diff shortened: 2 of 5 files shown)


mmckegg:
@mixmix 

I really like the way the blog previews look, however, I'm not a fan of the expanding in the main feed view. I think that clicking the preview should instead go to the message view which should have the full expanded blog post.

My reason for that is I'm trying to keep the feed views as stateless as possible (and avoid needing to updatethem with mutant). I want them be static read-only sprawls straight out of the DB. 

Also, small issue I noticed (with the current expanding style) is that the max-width on posts is causing the bottom of the blog post to get clipped. Might take a little bit of hacking to make it work since the blog is loaded async (moving to load the blogs only on the message view and as part of the loading will probably fix this, since the "View more..." link should then display correctly).

Changing the way that the message view works might be a little bit fiddly. I'm happy to help (or do the work) if you want 😺

mixmix:
yeah fine for the card to link through to expanded view.
Have you got a solution for this? we hacked something together for opening a gatherings object in expanded view in Patchbay but it's not that tidy. I've since added an async router in Patchbay which is capable of loading a full message from just an id link. This means you can then pivot on the message type and provide a different view.

mmm. might be easiest to discuss on the phone?

mmckegg:
@mixmix 

I think it could just be the messages ID that it links to. So the message view is always the expanded blog post.

mixmix:
DONE!

modules/feed/html/rollup.js
```diff
@@ -244,6 +244,7 @@ exports.create = function (api) {
       var renderedMessage = api.message.html.render(item, {
         compact: compactFilter(item),
         includeForks: false, // this is a root message, so forks are already displayed as replies
+        renderAsCard: true,
```
mixmix: this is the only contentious bit IMO - it's the option to pass to all the items to let them know to be cards in the rolled up view ...`renderAsCard` is not perfect but it is descriptive ... maybe it could be `renderAsCardIfCan` ergg 
# :nauseated_face:

mmckegg:
@mixmix 

I had a play around with this and I suggested some changes in #767. My idea moves the resolution of blobs into patchcore `message.obs.thread`. 

The advantage to doing it here is that the view is no longer doing a database lookup fetch, and instead this is being handled my the model (thread). The view decides whether to render card or full based on whether the body content is available (msg.body).

This fixes the "View more..." problem, and now the global "Loading..." screen waits for the blob to resolve before displaying the post. It also means that the blob doesn't have to be resolved until the blog is actually clicked (not sure if that is a good thing or not -- maybe it should still fetch it).

How do you feel about this?

mmckegg:
p.s. I would be happy to merge #767 as is! So if you are happy, I'll merge it and push out a new release tomorrow!

mmckegg:
@mixmix 

I have merged my PR into this one. About to merge to master!

PR #767, opened 2018-04-10
suggestions for #763 add blog rendering

mmckegg:
This is a slightly different approach to #763 that instead of resolving the blog in the renderer, does it in patchcore in message.obs.thread.

The renderer decides whether to render as a card or full based on if the body content is available.

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
     "mutant-pull-reduce": "^1.1.0",
     "obv": "0.0.1",
     "patch-settings": "~1.1.0",
-    "patchcore": "^1.23.6",
+    "patchcore": "github:ssbc/patchcore#blogs",
     "pull-abortable": "^4.1.0",
     "pull-cat": "^1.1.11",
     "pull-defer": "^0.2.2",
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
     "mutant-pull-reduce": "^1.1.0",
     "obv": "0.0.1",
     "patch-settings": "~1.1.0",
-    "patchcore": "github:ssbc/patchcore#blogs",
+    "patchcore": "~1.24.0",
     "pull-abortable": "^4.1.0",
     "pull-cat": "^1.1.11",
     "pull-defer": "^0.2.2",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -43,7 +43,7 @@
     "mutant-pull-reduce": "^1.1.0",
     "obv": "0.0.1",
     "patch-settings": "~1.1.0",
-    "patchcore": "^1.23.6",
+    "patchcore": "github:ssbc/patchcore#blogs",
```
mixmix: I assume a new `patchcore` would be published and this would be modified before merging

plugs/message/html/render/blog.js
```diff
@@ -92,3 +91,7 @@ function BlogCard ({ blog, blobUrl, onClick, color }) {
 
   return b
 }
+
+function canRender (msg) {
+  return msg.value && msg.value.content && msg.value.content.type === 'blog' && isBlog(msg)
```
mixmix: oh, this isn't needed - `isBlog` is _tight_ as bro - it's built with json-schema validation, so you don't need to write all that paranoid code :P

plugs/message/html/render/blog.js
```diff
@@ -16,21 +19,34 @@ exports.needs = nest({
...
-    if (!isBlog(msg)) return
+    if (!canRender(msg)) return
 
-    var blog = Blog(api.sbot.obs.connection).obs.get(msg)
-    var content = opts.inRollup
-      ? BlogCard({
-        blog,
+    var content = null
+
+    // only render the message body if it is available (only in thread view), otherwise show card
```
mixmix: I'd write this comment with a little more context

```js
// show a card (if there's no body loaded) or the full blog (if the blog body is loaded)
// msg is decorated with a `body` attribute for the thread view
```

package.json
mmckegg: yep!

plugs/message/html/render/blog.js
mmckegg: haha! The reason I added it was because I was debugging something. But it does need to expose the isBlog function as message.html.canRender for deciding whether to show the post in "show x updates".

PR #789, opened 2018-05-02
Update NPM dependencies

christianbundy:
My original intention was to update _everything_ so we didn't have to deal with https://github.com/Level/leveldown/issues/455, but I abstained from two updates:

- ~~**flumeview-level**: ^2.0.3 → ^3.0.2 (no posts showing up in public feed, although they work elsewhere)~~
- ~~**scuttlebot**: ^11.2.1 → ^11.3.0 (scared of https://github.com/ssbc/scuttlebot/issues/493 and https://github.com/ssbc/scuttlebot/issues/495~~

I haven't noticed any changes, but this should help with some of the dependency gridlock issues.


christianbundy:
I added updates for both `ssb-private` and `ssb-backlinks`, which solved the issue with `scuttlebot`, so that's updated as well. The only outdated package is `flumeview-level`, which, when updated, causes the "Public" feed to show up completely blank (although all other feeds seem to work fine).

In any case, everything else seems to be working great!

mmckegg:
Cool! Thanks for this @christianbundy. Will take a look, and see if I can figure out what's up with that last dependency.

mmckegg:
Looks like the frameless mode isn't working on macOS. Must be something else changed in electron@2?

screen shot 2018-05-03 at 11 11 32 am

mmckegg:
Also the more menu is no longer positioning under the button when clicked:

screen shot 2018-05-03 at 11 13 10 am

Should be like this:

screen shot 2018-05-03 at 11 14 12 am

mmckegg:
Looking through the package-lock, it looks like there are a couple of dependencies that still need to be updated to flumeview-level@3 to completely eliminate the level@1.7 dependency.

- secure-scuttlebutt
- flumeview-search

---

p.s. sorry about the torrent of comments! Posting as I go 😆

mmckegg:
@christianbundy I managed to get `flumeview-level@3` working in patchwork by bumping the index version on `roots.js`.

```diff
diff --git a/sbot/roots.js b/sbot/roots.js
index 643bdd01..6fe106f6 100644
--- a/sbot/roots.js
+++ b/sbot/roots.js
@@ -12,7 +12,7 @@ var getRoot = require('patchcore/message/sync/root').create().message.sync.root
 var getTimestamp = require('patchcore/message/sync/timestamp').create().message.sync.timestamp
 
 module.exports = function (ssb, config) {
-  var create = FlumeViewLevel(2, function (msg, seq) {
+  var create = FlumeViewLevel(3, function (msg, seq) {
     var result = [
       [getTimestamp(msg), getRoot(msg) || msg.key]
     ]

diff --git a/package.json b/package.json
index b646f8df..4a8e1c64 100644
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
     "emojilib": "^2.2.9",
     "fix-path": "^2.1.0",
     "flatpickr": "^4.4.4",
-    "flumeview-level": "^2.0.3",
+    "flumeview-level": "^3.0.2",
     "flumeview-reduce": "^1.3.13",
     "flumeview-search": "github:mmckegg/flumeview-search#fix-hang-on-short-words",
     "hashlru": "^2.2.0",
```

christianbundy:
Thanks for taking a look at this and provide input! I'm pretty busy with work but my goal is to find some time over the next few days, but I want to acknowledge your speedy review. :+1:

christianbundy:
Thanks for the tip on `secure-scuttlebutt` and `flumeview-level`, I'll see whether I can squeeze those fixes in too. As for this PR:

- I learned that the first argument to `FlumeViewLevel` matters, and that's all working correctly.
- I configured the window frame to be [`hidden`](https://github.com/electron/electron/blob/master/docs/api/frameless-window.md#hidden) but I'm not on macOS -- would you mind double-checking to make sure it looks right on your system?
- I replaced the old menu popup code with the [new `menu.popup` method signature](https://github.com/electron/electron/blob/master/docs/api/menu.md#menupopupoptions) and the popup is in the correct location.
- I replaced checks for `(typeof value.content === 'string')` with `(value.private)` and there are no more private messages showing up in public feeds.

Please let me know if you notice anything else!

mmckegg:
Oh, the one theory I do have is that it occurs when a thread has a new reply or like downloaded. This seemed to be the case for all the times I experienced it.

mmckegg:
So it does seem to be something to do with backlinks doubling up on items. Just noticed this in my feed:

screen shot 2018-05-04 at 11 49 49 am

Note how the same reply post has repeated twice. Sure enough, when I tried to view the thread, I got the infinite loading bug. Here's where these are handled in patchcore: https://github.com/ssbc/patchcore/blob/master/backlinks/obs.js#L57

The same backlinks query gets cached for likes, replies, and thread. My guess is that the ssb-backlinks plugin is returning the same object multiple times (for some reason). I just remembered that @arj03 just made some updates to ssb-backlinks to get it working with the latest sbot. Maybe it could have been removing the `{index}` option.

mmckegg:
Of course, one easy fix would be to enforce uniqueness when appending backlinks to the collection.

mmckegg:
Okay, I found out what is causing the double up (and infinite loading) problem!

https://github.com/ssbc/patchwork/blob/master/sbot/live-backlinks.js#L13

For some reason `backlinks.read({old: false})` is not working. Instead of only returning a live stream of new backlinks, it streams every single backlink in the entire database!! 😢 

It will only cause the post to double up if it is currently on screen in patchwork at the time that the stream finally gets up to your message. Since the message will be near the end, everything seems to work for a while at first and then it hits.

mmckegg:
@christianbundy 

I managed fixed the post [double ups / infinite loading](https://github.com/ssbc/patchwork/pull/789#issuecomment-386471072) by making a [PR to flumeview-query](https://github.com/flumedb/flumeview-query/pull/12).

If you want to get it working in Patchwork before this is merged:

```bash
cd patchwork
npm install mmckegg/flumeview-query#add-index-option
```

```diff
diff --git a/package.json b/package.json
index 0213d366..92e854c7 100644
--- a/package.json
+++ b/package.json
@@ -30,6 +30,7 @@
     "fix-path": "^2.1.0",
     "flatpickr": "^4.4.4",
     "flumeview-level": "^3.0.2",
+    "flumeview-query": "github:mmckegg/flumeview-query#add-index-option",
     "flumeview-reduce": "^1.3.13",
     "flumeview-search": "github:mmckegg/flumeview-search#fix-hang-on-short-words",
     "hashlru": "^2.2.1",
```

christianbundy:
That looks great, solid debugging! Would you prefer I added that to this branch, or should I wait for an official release from `flumeview-query`?

I've just pushed another commit to work on the window frame regression -- my new line of thinking is that changing `lib/window.js` was misguided, that the real issue was the fact that [`hidden-inset` was deprecated in favor of `hiddenInset`](https://github.com/electron/electron/pull/11975), although I don't remember ever seeing the deprecation warnings. The window looks the same for me regardless (useless screenshot attached) but I'm hoping this will fix the issue on macOS. :crossed_fingers: 

screenshot from 2018-05-04 11-02-22

christianbundy:
@mmckegg Anything I should update on this branch? I think all of the merge conflicts should be resolved now.

mmckegg:
@christianbundy I'm gonna try this out and see how it goes!

christianbundy:
:raised_hands:

PR #802, opened 2018-05-11
Add forward and back keyboard shortcuts and menu bar

mrchrisadams:
Hi there.

I started using Patchwork last week, and I really found myself missing a way to trigger back and forward buttons.

So I've had a go at adding these - first looking into doing this with the Menu and [MenuItem](https://electronjs.org/docs/api/menu-item#menuitemclick) Electron APIs,  then reverting to using mousetrap to add the keyboard shortcuts in the app itself.

With this branch, you can go back and forward with `cmd+[` and `cmd+]` respectively, or do the same with `cmd+left` or `cmd+right`.

However, I think I need some pointers on:

- making this work cross platform 
- how best to integrate this to the rest of the codebase

### More details

The Electron APIs seem like the better way to go for adding keyboard shortcuts, as it makes shortcuts much more discoverable and accessible (they show up when searching in help, and the shortcuts will show up in the menu bars, in the UI).

screen shot 2018-05-11 at 11 40 10

However, we define the menu in the root `index.js` file, which seems to set up a shell for behaviour we specify in `main-window.js`, with the `views.goForward()` and `views.goBack()` functions in the _inner_ buttons in the window chrome.

screen shot 2018-05-11 at 11 11 58 

So, here's what I think I want to do, in the root `index.js` file.

```javascript
   menu.splice(4, 0, {
      label: "History",
      submenu: [
        {
          label: "Forward",
          accelerator: "CmdOrCtrl+]",
          click: () => {
            console.log("go Forward", views.goForward)
            views.goForward
          }
        },
        {
          label: "Back",
          accelerator: "CmdOrCtrl+[",
          click: () => {
            console.log("go Forward", views.goBack)
            views.goBack
          }
        }
      ]
    });
```

I don't understand how I access `views.goBack` available here though, as it seems to be using depject, which is new to me.

How should I be making `views.goBack` in the electron menu, so we have keyboard short cuts here?

#### Hopefully useful background links:

More on adding keyboard shortcuts for electron:
https://electronjs.org/docs/tutorial/keyboard-shortcuts

More on making them appear in the menu
https://electronjs.org/docs/api/menu
https://electronjs.org/docs/api/menu-item


mrchrisadams:
Okay, after spending a bit of time looking into this with @ciex, I'm now wondering if sending events using [this pattern in patchbay](https://github.com/ssbc/patchbay/blob/master/index.js#L42) is a better approach:

```
{
      label: 'Select Next Tab',
      accelerator: 'CmdOrCtrl+Shift+]',
      click() {
        windows.main.webContents.send('nextTab')
}
```

Presumably, we might be able listen to the `nextTab` event elsewhere in the application.

mmckegg:
@mrchrisadams 

Yes, using [electron ipc](https://electronjs.org/docs/api/ipc-renderer) is the way I would recommend handling this. You could then add something like this to `main-window.js`:

```js
electron.ipcRenderer.on('goBack', views.goBack)
electron.ipcRenderer.on('goForward', views.goForward)
```

mrchrisadams:
Okay. thanks @mmckegg, that makes things much clearer, and pretty much exactly what we needed 👍 

I've updated this to use the Electron API now, so that:

- you can move back and forth by using the cmd/ctrl and square brackets - (this is the convention used by Firefox, Chrome, and [tools with browsers built in like Dash](https://kapeli.com/dash_ios)).
- these keyboard shortcuts appear in the new 'History' menu

### Summary of changes

I had to make `openMainWindow` return a [BrowserWindow][browserWindow] instance so I could access the `webContents` object in order to send messages with the [send method on it](https://electronjs.org/docs/api/web-contents#contentssendchannel-arg1-arg2-)

I'm not aware of any other effects this might have - please let me know if there's a better way to get a reference to the [BrowserWindow object][browserWindow], and I'll update the PR accordingly.

[browserWindow]:https://electronjs.org/docs/api/browser-window

If this looks okay, I'll update the name of this PR, so it's in shape to merge in.

On a related note - I didn't see an obvious place to update the documentation as we learned about the codebase. [Scuttlebutt.nz](https://www.scuttlebutt.nz/modules.html) seemed like the first place, but when after reading through it, it looks like it's concerned about the protocol, not Patchwork the app itself.

If you let me know, I'll some high level docs as part of a separate PR.

mmckegg:
@mrchrisadams 

>  I didn't see an obvious place to update the documentation as we learned about the codebase. Scuttlebutt.nz seemed like the first place, but when after reading through it, it looks like it's concerned about the protocol, not Patchwork the app itself.

There isn't really much in the way of that right now. I did start [a contributing guide](https://github.com/ssbc/patchwork/blob/master/contributing.md), but it doesn't have much in it right now. That could be a good place for some things. Feel free to start another markdown file that can be linked from the readme!

mmckegg:
Thanks for the PR @mrchrisadams, merged! Will be in the next release.

The only change I made was to swap the ordering of the Back and Forward menu items to match the order that Google Chrome uses.

PR #807, opened 2018-05-21
Add name publicness note

neurodynamic:
I think it's important to have this be explicit.

I tried working on adding a similar note for assigning profile images, but had some trouble figuring out maintaining the styling after putting the file chooser behind a "Confirm" message. I think it'd be good to have one for that, too, though.


mmckegg:
Thanks @neurodynamic! I modified this to show the warning as an h3 instead of h2. But otherwise, this will be in the next release!

PR #812, opened 2018-06-04
update sbot 11.3.0 + new plugins

mixmix:
These are all now in master in Patchbay + Ticktack.
The problem is that when I open Patchwork now, it re-indexes all these plugins because of the different versions we're running. It makes switching btwn Patchwork and these other apps pretty much intolerable.

I think this is a broader conversation that we might need to have - as in what's our strategy here in the short and long term

:heart: :cactus:


cryptix:
Ran this through the whole day and worked fine afaict.

mixmix:
I think there are now newer sbot versions out.

mmckegg:
@mixmix I think that the recently merged #789 has updated these. Closing!

mixmix:
Cool!

Issue #810, opened 2018-05-29
Change "Publish" button label to "Preview and publish"


ivilata:
Since messages in SSB cannot be edited so the user need be extra careful when posting, and hitting "Publish" actually shows you a preview dialog before really posting, it would be nice that the button label was explicit about this (e.g. by changing it to "Preview and publish") so that the user knows that hitting the button doesn't post the message straight away and gives an opportunity for preview (so that one doesn't need to prepare and preview the message in an external tool).

mmckegg:
Implemented in #817 and #818. Closing!

PR #817, opened 2018-06-15
Preview and publish

stripedpajamas:
To address #810

I think this tiny change will clear up a lot... it took a couple mistakes for me to realize that profile publishes and gathering publishes don't get previewed, but public/private posts do. This should make that more explicit.

If anyone knows any of the supported locales and wants to update those files that would be great. I only know English so that's what we've got here.


mmckegg:
Great thanks @stripedpajamas! Yes, we'll need to get those locales updated before this can be released.

PR #820, opened 2018-06-16
Add appdata and desktop files

ptomato:
Hi, I'm trying to package Patchwork for distribution on Flathub (https://flathub.org) and I thought you might like to have these metadata files as they could be useful for the AppImage build as well. These files are used for providing metadata (descriptions, icons,...) for various app stores on Linux.

### DIFF
--- a/assets/nz.scuttlebutt.Patchwork.appdata.xml
+++ b/assets/nz.scuttlebutt.Patchwork.appdata.xml
@@ -1,0 +1,60 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop">
+  <id>nz.scuttlebutt.Patchwork</id>
+  <name>Patchwork</name>
+  <project_license>AGPL-3.0</project_license>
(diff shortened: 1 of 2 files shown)


mmckegg:
This looks fine @ptomato. The only issue I can see with merging this is keeping the release info up to date: https://github.com/ptomato/patchwork/blob/535973ce3eb49d76b9e640991bbf38357bd61ae0/assets/nz.scuttlebutt.Patchwork.appdata.xml#L31-L55

ptomato:
Yeah, that seems like (a small amount of) annoying manual work that would have to be done each release. Do you think it should just be removed instead?

mmckegg:
@ptomato yes, probably should just be removed. Unless you want to create a script that pulls this out of the releases tab?

ptomato:
OK, here it is with that section removed.

mmckegg:
Okay, thanks @ptomato, merging!

Issue #822, opened 2018-06-26
Is that possible to have the installation of SSB/patchwork in a USB thumbdrive?


deepcell:
For example, my fedora is installed in a thumbdrive, is that possible to make the installation of SSB/patchwork in another thumbdrive (I want have it dedicated with a 64gb usb drive)? It would be very nice if this was an option.
Thanks for any idea.

Serkan-devel:
I guess if one places the `.ssb` on the drive and configure patchwork to run from there, it could work

deepcell:
I will try that, just to say in my Fedora it is also in my .config/
folder, then I have it in 3 different locations:

~/.config/Patchwork/ (it includes cache, local storage, cookies,
preferences, dict..)
~/.ssb/
~/Downloads/Patchwork-3.9.0-linux-x86_64.AppImage (executable)

I just move it all to the usb drive (size about 700mb)

Then I executed Patchwork and I think I messed up, because it recreate
all the folder I  mentioned above and it is empty. The Patchwork App
there is nothing related to data source, paths, etc.. perhaps I have
to check the source code and compile myself.

Any idea which file is related to configuration?

On 6/26/18, Serkan-devel <notifications@github.com> wrote:
> I guess if one places the `.ssb` on the drive and configure patchwork to run
> from there, it could work
>
> --
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub:
> https://github.com/ssbc/patchwork/issues/822#issuecomment-400343176


-- 
PGP Key ID: 0x3F75FA35

stripedpajamas:
@deepcell the problem is that Patchwork looks in `~/.ssb` for your secret+db, and if it doesn't find it, it initializes fresh. Patchwork pulls in [ssb-config](https://github.com/ssbc/ssb-config) and that allows a path override. right now, the config overrides aren't configurable in Patchwork -- [this line](https://github.com/ssbc/patchwork/blob/master/index.js#L46) specifies a `server` override but nothing else.

@mmckegg I could add a command line option (e.g. `--ssb-dir <dir>`) to pass into `ssb-config` to support USB Patchworkers?

deepcell:
@stripedpajamas I downloaded a fresh new copy and apply your changes [](https://github.com/ssbc/patchwork/pull/829) to the files, then try to run again calling with the parameters, but it still looking the same .ssb source. 

[...]

PR #829, opened 2018-07-09
Add support for custom ssb path

stripedpajamas:
This should allow stuff like running from a USB stick (#822). I added `commander` as a dependency so other command line options could be added later if the need arises. 

With this PR, someone could set their shortcut to something like `patchwork --path /custom/path` or even `patchwork --path .` to use the cwd, which should be perfect for portable installs.

Since git-ssb doesn't take in configs, I have Patchwork skip starting it up if a custom path was passed in. If git-ssb changes to allow more granular configs, we can take that conditional out.

### DIFF
--- a/index.js
+++ b/index.js
@@ -12,19 +12,12 @@ var WindowState = require('electron-window-state')
 var Menu = electron.Menu
 var extend = require('xtend')
 var ssbKeys = require('ssb-keys')
-var program = require('commander')
 
 var windows = {
   dialogs: new Set()
 }
 var ssbConfig = null
 var quitting = false
-
-// parse program args
-program
-  .option('-g, --use-global-ssb', 'Use a global instance of sbot (Patchwork starts its own instance by default)')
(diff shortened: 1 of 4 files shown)


mmckegg:
@stripedpajamas 

This looks pretty good, and I'm tempted to accept as is. The only issue I have is that ssb-config actually has it's own command line parser ([minimist](https://github.com/substack/minimist) via [rc](https://github.com/dominictarr/rc) which is clever, but I'm not a huge fan of) that that takes the args passed to patchwork and turns them into config automatically.

So passing `--path SOMETHING` to patchwork would actually already set `config.path: "SOMETHING"`. You can also do things like `--friends.hops 10` which would set `config.friends.hops`.

It might be worth having a play around with the built in arg parser and see if you can get it do what you want. If not, I'll merge this!

stripedpajamas:
Nice catch!! It works perfectly. I updated my PR to just not start git-ssb if a custom path was passed in, so it doesn't blow up.

PR #831, opened 2018-07-10
Bring NPM dependencies up-to-date

christianbundy:
I'd imagine doing this every so often is probably going to be much less painful than trying to do heaps of them when some downstream dependency breaks.

- patchcore       (1.27.2 -> 1.28.0)
- ssb-backlinks   (0.7.1 -> 0.7.2)
- ssb-search      (1.1.1 -> 1.1.2)
- app-root-path   (2.0.1 -> 2.1.0)
- electron        (2.0.3 -> 2.0.4)
- flatpickr       (4.5.0 -> 4.5.1)
- moment-timezone (0.5.20 -> 0.5.21)

Please let me know if you'd prefer me to work on this another way -- I usually run the edge of `master` so it's easy for me to progressively update our dependencies and make sure everything keeps chugging along.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -43,9 +43,9 @@
       }
     },
     "@types/node": {
-      "version": "8.10.20",
-      "resolved": "https://registry.npmjs.org/@types/node/-/node-8.10.20.tgz",
-      "integrity": "sha512-M7x8+5D1k/CuA6jhiwuSCmE8sbUWJF0wYsjcig9WrXvwUI5ArEoUBdOXpV4JcEMrLp02/QbDjw+kI+vQeKyQgg==",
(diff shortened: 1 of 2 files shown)


mmckegg:
According to the package-lock, we're already running these versions, so I'm just gonna merge this!

christianbundy:
I think that's only when you use `npm install`, but when downstream packagers ([example](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=ssb-patchwork#n26)) install Patchwork's dependencies we'll generally use `npm ci` so `package.json` is ignored and we only install whatever's in the `package-lock.json` at the time of release.

PR #850, opened 2018-08-29
Extract shared styles from light and dark themes

christianbundy:
This takes all of the shared MCSS styles from the light and dark themes and places them in a directory called `base`. This replaces the original theme directories with *only* the diffs from the shared MCSS styles, in an effort to reduce duplication and keep things DRY.

This is a mechanical edit from [an experimental MCSS serializer](https://github.com/mmckegg/micro-css/pull/3) so it's currently clobbering whitespace and (unnecessary) semi-colons. I'd love any input, although I'm specifically interested in:

- theme regressions (e.g. wrong background color)
- source code regressions (e.g. blocks with no rules like `Document { p { span {} } }`)
- any sort of suggestions for improvement (please!)

My work is *actually* being done in [the `diff-themes` branch](https://github.com/ssbc/patchwork/tree/diff-themes/styles), but I've had a bunch of iterations (with huge diffs) so this PR is coming from a new branch with only a single non-nasty commit.

---

After noticing that the themes had diverged on silly CSS ambiguities, I wrote a [hacky color normalizer](https://github.com/ssbc/patchwork/blob/7fcce7f3fabe4921e89cbffafdcca460f4d6879a/styles/serializer.js#L61-L93) with three cascading policies:

1. If possible, use a color name (e.g. `white`)
2. If possible, use a shorthand hex (e.g. `#aaa`)
3. If necessary, convert to lowercase (e.g. `#d77986`)

It's equally easy to prefer hex representations over color names, but it felt silly to have the base theme specify `#fff` and have the light theme "diverge" and specify `white` so I wanted to make sure we have *some* policy. I don't have any strong opinions either way. I'd love to hear if anyone would rather:

- prefer hex representations to color names
- prefer uppercase to lowercase
- prefer 6-digit hex to 3-digit shorthand


neale:
I ran this PR for a few hours today and didn't have any issues. 
I test both themes with different font sizes and nothing broke. It could probably benefit from a more thorough test, but I didn't have issues with elements or spacing.

mmckegg:
Nice work @christianbundy! I will give it a try and proper review soon.

mixmix:
hey @christianbundy the next question on my mind is: 
- should this be merged before or after the epics open (tags, polls)? 

The original idea was to simplify maintenance and hopefully minimise need for people to do styling e.g. a dark theme when they make a feature. Do you think it would be easy to run the process on features coming in?

christianbundy:
@mixmix 

> Do you think it would be easy to run the process on features coming in?

You got it. The whole process is trivial and can be done from any commit. I'll include my workflow at the bottom, which is mostly just Git work to keep this PR branch clean. The `node extract.js` bit is the only part that matters and it's completely automatic.

**I think it would be simplest to merge this PR last**, otherwise we'd need to run the theme extractor on all of the outstanding PR branches to merge them cleanly. We *could* merge this PR first, but it seems simpler to use the extractor once rather than having to use it on every open PR.

---

```bash
git checkout diff-themes \
&& node extract.js \
&& git add base diff-dark diff-light  \
&& git commit -S \
&& git reset --hard \
&& git checkout diff-themes-mcss \
&& git checkout diff-themes -- base diff-light diff-dark \
&& rm -rf light dark \
&& mv diff-light light \
&& mv diff-dark dark \
&& git add base light dark \
&& git commit -S --amend
```

mixmix:
Ok good to know. I think this will still help us merge earlier because
we'll know some dark styling might not need to be done.

mmckegg:
@mixmix gonna pull this in now and take a wee look!

styles/base/index.js
```diff
@@ -0,0 +1,20 @@
...
+  if (/\.mcss$/i.test(file)) {
+    result += fs.readFileSync(path.resolve(__dirname, file), 'utf8') + '\n'
+  }
+
+  if (/\.css$/i.test(file)) {
+    additional += fs.readFileSync(path.resolve(__dirname, file), 'utf8') + '\n'
+  }
+})
+
+const flatpickr = require.resolve('flatpickr/dist/flatpickr.css')
```
mmckegg: I noticed that flatpickr styles are included in the `base` css as well as `dark` and `light`. Should just be in the `base`.

styles/index.js
```diff
@@ -1,4 +1,6 @@
+const base = require('./base')
+
 module.exports = {
-  light: require('./light'),
-  dark: require('./dark')
+  light: base + require('./light'),
+  dark: base + require('./dark'),
```
mmckegg: I would prefer if the base styles were included in the theme's `index.js` instead of the shared one here. 

To me, this would seem easier to follow.

styles/base/profile-list.mcss
```diff
@@ -0,0 +1,105 @@
...
+        }
+      }
+    }
+    -following {
+      background-image: svg(following)
+    }
+    -connected {
+      background-image: svg(connected)
+    }
+    @svg connected {
```
mmckegg: The only background issue I've noticed is that the _connected_ indicator isn't showing up on the "Connected Pubs" list.

screen shot 2018-09-24 at 2 12 41 pm

Whereas in the last release it looked like this:

screen shot 2018-09-24 at 2 33 00 pm

I tracked this down to the way that MCSS handles the embedded SVGs. They don't cascade in the same way as css stuff. This is because they aren't really css at all, but helpers for compiling data urls. The data url gets compiled at the point you call `mcss.compile` and no more cascading can happen after this.

I managed to fix this by compiling the light/dark with the base theme as a single mcss call. The only alternative is to move all of the SVG stuff into light/dark or base, but that wouldn't be as nice.

Here's the quick and dirty approach I tested:

```diff
diff --git a/styles/light/index.js b/styles/light/index.js
index 6a83b7b4..7ba223df 100644
--- a/styles/light/index.js
+++ b/styles/light/index.js
@@ -4,6 +4,16 @@ var compile = require('micro-css')
 var result = ''
 var additional = ''
 
+var basePath = path.join(__dirname, '..', 'base')
+fs.readdirSync(basePath).forEach(function (file) {
+  if (/\.mcss$/i.test(file)) {
+    result += fs.readFileSync(path.resolve(basePath, file), 'utf8') + '\n'
+  }
+  if (/\.css$/i.test(file)) {
+    additional += fs.readFileSync(path.resolve(basePath, file), 'utf8') + '\n'
+  }
+})
+
 fs.readdirSync(__dirname).forEach(function (file) {
   if (/\.mcss$/i.test(file)) {
     result += fs.readFileSync(path.resolve(__dirname, file), 'utf8') + '\n'
```
And then removing the appending of base in `styles/index.js`.

I'm sure there is a tidier way to handle this, but hopefully this gives you some ideas!

styles/index.js
mmckegg: After commenting this, I discovered why some of the SVG images weren't appearing correctly (see other comment), so I think this will have to be changed to fix that problem.

christianbundy:
Thanks for the review @mmckegg! I think that's a great suggestion, I've just refactored both `index.js` files to use arrays and `forEach()` methods to handle the files in both directories. I've also switched to `push()`ing the `results` and `additional` and joining with a newline once at the end rather than doing a string mutation for each fie.

The SVG looks to be working correctly on my end, please let me know if you notice any other funky behavior!

styles/base/index.js
christianbundy: Done, removed from dark and light. :+1:

styles/base/profile-list.mcss
christianbundy: Really appreciate this code, it made resolving the issue *much* easier.

styles/base/index.js
mmckegg: Oh, these flatpickr styles will now be needed in the `light` and `dark` themes now that you are no longer merging this file in. In fact base/index.js can now be deleted.

It is used by the gatherings editor when clicking the date field. Without these additional styles, the date picker is not visible.

styles/dark/markdown.mcss
```diff
@@ -1,111 +1,51 @@
...
-  (a.bad) {
-    color: inherit
-    :hover {
-      text-decoration: inherit
-    }
-  }
-
-  word-break: break-word
-  (img) {
+  img {
```
mmckegg: I've noticed that the theme files are missing _parentheses_ around the selector, but the ones in `base` are fine. I think that (due to a bug in mcss) if any instance of that rule is defined with _parentheses_, that all are treated this way. So it doesn't seem to be breaking anything in the patchwork styles. However, for consistency (and in case the mcss bug is fixed), these should be updated to have the _parentheses_ on all of these selectors.

This is handled in the tokenizer by `deep: true`. This is for choosing when not to use the direct descendent selector (`>`). Which I see is being handled by [the serializer](https://github.com/mmckegg/micro-css/pull/3/files#diff-72eec7d9bc03ee5b16aa5768cb53d614R46). Maybe there is a bug in your diffing?

christianbundy: I think this might be a serializer bug where `deep: true` is being treated like `color: "red"` or something: it's preserved in the base because it's found in both token objects, but removed from the diffs because it's treated like a shared property. I think it's diffing on:

``` css
A {
  b {
    background: blue
    color: green
    deep: true
  }
}

Comment (christianbundy) [reply styles/dark/markdown.mcss]: Turns out the `deep` issue wasn't a serializer bug, it was the diff function that wasn't passing `deep` along in the diff. I couldn't find an elegant solution I liked so I hacked together a [a solution that works](https://github.com/ssbc/patchwork/commit/c20552d7246784403f17117ed9cc775d71ff1738), which basically says:

```
if (foo.deep === true) {
  foo.deep = 1
}
```

This means that it passes our two requirements:

- it's technically different, so it's passed along to the serializer
- it's still truthy, so the serializer treats it the same as before and prints `(foo)`

styles/base/index.js
christianbundy: Ha! Thanks, I didn't really consider how useless `base/index.js` would be with this solution. Should be resolved now.

mmckegg:
Okay, this is merged! Thanks for this @christianbundy. Will make theme stuff much easier.

christianbundy:
Sweet, thanks for iterating with me and reviewing my code. Please ping me if this breaks anything. :fire:

PR #856, opened 2018-09-26
add support for links to private message with unbox key

ahdinosaur:
depends on

- https://github.com/ssbc/secure-scuttlebutt/pull/220 (which depends on latest `ssb-keys` and `ssb-config`)
- https://github.com/ssbc/ssb-ref/pull/19
- https://github.com/ssbc/ssb-ref/pull/20

an example message with a private link is `%37lW2zSbuFkBBVYYCR9Zjyve4YBFbTTmEw2KfYJVEm8=.sha256`

with the private link of `%WgVG9T2IryRoPMCQk7znuMt2Cmo/shgnrbn0wY6gc3M=.sha256?unbox=AZlrtZIJQiHqgwCaB0GgtIiFXha+XN5y6n5NJz/HtunP`

this change allows you to navigate to such a link, without the compose input (because you aren't included in that private conversation).


mmckegg:
@ahdinosaur 

Had a look through the code and the only thing that I'm worried about is turning the `id` into an object. Pretty sure this is fine, but we can really only find out with use.

I'm gonna go ahead and merge this, then run for a bit and see how it goes!

mmckegg:
@ahdinosaur A couple of follow up ideas:

- right click + copy link reference should probably include the unbox key
- make it clearer that you are viewing the thread with a special key in message view

Issue #838, opened 2018-07-24
Limit on number of recipients silently fails and freezes "private" window.


Zaijo:
Hello there. Patchwork 3.10.1 private edit window freezes when trying to mention more than 7 recipients in a private message. Why the limit?

How to reproduce:
- Write a private message to 8 recipients.
- Check error log in Electron console.
- You get: `Error: max recipients is:7 found:8`

First thing is. **Do we need the limit?** Why is it there? Does it have something to do with the architecture of ssb?

If we need the limit, I'd like to see a warning i.e.: "Patchwork only allows 7 recipients in a private message, because <reason follows>."

cryptix:
The limit is arbitrary but not unwarranted. You might want to consult the [private messages section of the protocol guide](https://ssbc.github.io/scuttlebutt-protocol-guide/) to learn more about how they function.

The gist is this, though: Each recipient will get an individual decryption key for a private message. Each of these adds 49bytes to a message. Apart from the size constraint (8kb per message) having lot's of recipients on private messages will also lead to lots of overhead/wasted CPU cycles for all the non-recipients of a messages. Since there are no `TO: ...` fields everybody tries to decrypt all the messages they replicate/feeds they follow. This is manageable with 7 or 8, maybe with 10 but not much higher, I fear. For larger groups we will need something like shared encryption keys.

Thanks for the report, though. Patchwork definitely shouldn't freeze over this.

mmckegg:
Yeah, this is a bug. We'll get it fixed for the next release.

mmckegg:
Closing as this has now been addressed in #859 and will be in the next release.

PR #859, opened 2018-09-28
Add catch for too many private recipients

stripedpajamas:
Addresses #838 

Looks like this (changes welcome):

too_many_recps


modules/message/sheet/preview.js
```diff
@@ -21,6 +21,24 @@ exports.create = function (api) {
       var isPrivate = msg.value.private
       var isRoot = !msg.value.content.root
       var recps = (msg.value.content.recps || []).filter(id => id !== msg.value.author)
+
+      // handle too many private recipients
+      if (isPrivate && recps.length > 7) {
+        return {
+          content: [
+            h('h2', ['Too many recipients']),
```
mmckegg: These strings should use `i18n` to make them localisable. It should just be a matter of wrapping with a call to `i18n`, for example:

```
h('h2', [i18n('Too many recipients')])
```

But you'll need to make sure i18n is included in scope (I think it is in this file).

modules/message/sheet/preview.js
```diff
@@ -21,6 +21,24 @@ exports.create = function (api) {
...
+            h('div.info', [
+              h('p', [
+                'Private messages can only be addressed to up to 7 people. ',
+                'Your message has ' + recps.length + ' recipients. '
+              ]),
+              h('p', ['Please go back and remove some of the recipients.'])
+            ])
+          ],
+          classList: ['-private'],
+          footer: [h('button -cancel', { 'ev-click': cancel }, i18n('Cancel'))]
```
mmckegg: I think this button "Cancel", should be called "Close".

Serkan-devel:
but why should recipes be limited?

stripedpajamas:
@Serkan-devel see #838 (specifically the response from cryptix) for more info. It is just part of the protocol.

PR #860, opened 2018-09-29
fix isParticipant to handle the case where recps includes a feed mention

ahdinosaur:
with the previous private link change, i disabled the compose when you aren't a participant.

i didn't realize `recps` could include a feed mention, this fixes that. :sweat_smile:


mmckegg:
@ahdinosaur  Whoops. Is this a big break? Should I push out a point fix?

PR #871, opened 2018-10-08
Update French translation

yannicka:


christianbundy:
I'm not sure how many changes were made, but it looks like it's a full-file diff because the indentation was changed from spaces to tabs. I think it might be best to replace the new tabs with their original indentation.

PR #883, opened 2018-11-01
Ensure that msg.value is truthy

### LINKED ISSUE #882: Opening new channel throws an error and hangs forever
I am running from source code `master, tag: v3.11.31`

Reproducing from my side is so straight-forward: Click `more channels`, then click on any channels. It hangs forever and this appears in the logs

```
TypeError: Cannot read property 'content' of undefined
at checkBump (/home/blabla/patchwork/sbot/channel-feed.js:124:17)
```

I logged the message object and it had no `value` property

christianbundy:
Resolves #882 

---

This resolves and edge-case where a message is being passed to checkBump. I'm unclear on *why* this is necessary, as sbot seems to be reporting that the message doesn't exist, but it should resolve the bug.

cc: @smowafy could you test this and let me know if it resolves the issue?

### DIFF
--- a/sbot/channel-feed.js
+++ b/sbot/channel-feed.js
@@ -116,6 +116,7 @@ exports.init = function (ssb, config) {
 }
 
 function checkBump (msg, { channel }) {
+  if (!msg.value) return
   if (msg.value.content.type === 'vote') return
   if (normalizeChannel(msg.value.content.channel) === channel) {
     if (getRoot(msg)) {


smowafy:
@christianbundy yea this works locally, but I want to make sure I understand correctly.

I opened `#software` channel now and I can't find the message we mentioned in the issue. So either I receive the message without the value option or it lost it along the way. Am I correct?

I'm not aware of the internals and I'm actually interested to know the root cause of this. Also thanks a lot for your time checking this issue!

christianbundy:
No problem! My best guess is that you don't have those messages because `@O71isksx77Du8JQvsf91sdPwW4kPnlZgnw8IotOYxlM=.ed25519` isn't someone within your follow range, but that you only have the *replies* to that message. Each reply specifies the parent message, so I think the client likely tried to render the parent message without actually checking to see if it's a message you have. For reference, here's how the thread looks for me:

reply

You're missing the root message, but my guess is that Dominic is within your follow range, so when Patchwork tried to render the thread it crashed when it assumed it had the entire thread. It could totally be something else, but that's the best explanation I can come up with.

cc: @mmckegg does that sound right? or is there a better way to do this?

PR #885, opened 2018-11-05
Fix inaccurate slovak localization

iplesko:
The original localization string "označil ťa v sieti" means "mentions **you** in your network" which doesn't make sense when rendered in Patchwork.


Zaijo:
Thanks. You fixed it correctly.

PR #886, opened 2018-11-05
WIP: Upgrade NPM dependencies

christianbundy:
This commit upgrades our NPM dependencies and upgrades Scuttlebot to 13.0.3, as well as using unix sockets when they're available. Untested on Windows, and will likely depend on https://gitlab.com/staltz/sbot-gossip/issues/1, so this PR currently uses the vanilla gossip plugin from Scuttlebot.

cc: @staltz please let me know if I can help get your gossip plugin working with sbot@13 :+1:


index.js
```diff
@@ -178,13 +178,23 @@ function setupContext (appName, opts, cb) {
...
+  if (process.platform === 'win32') {
+    // fix offline on windows by specifying 127.0.0.1 instead of localhost (default)
+    ssbConfig.remote = `net:127.0.0.1:${config.port}~shs:${pubkey}`
+  } else {
+    const socketPath = Path.join(ssbConfig.path, 'socket')
+    ssbConfig.connections.incoming.unix = [{ 'scope': 'local', 'transform': 'noauth' }]
+    ssbConfig.remote = `unix:${socketPath}:~noauth:${pubkey}`
+  }
+
+  console.log(ssbConfig)
```
christianbundy: Shamelessly stolen from https://github.com/ssbc/patchbay/pull/264

staltz:
@christianbundy I would probably move everything dht-invite related to a branch, and take careful steps forward. `@staltz/sbot-gossip` is clearly a hack, I don't intend it to be good software, so it comes "without warranty". I'll comment on that error in specific in that issue. The good way forwards is to fund the rewrite of a new gossip plugin.

mmckegg:
@staltz are you suggesting that it would be better not to release the DHT invites in patchwork at this time?

staltz:
@mmckegg I'd love to see DHT invites in Patchwork rather sooner than later, but I don't have time at the moment to invest into fixing compatibility between `@staltz/sbot-gossip` and sbot 13. It's probably not too hard either, but currently I'm full of a bunch of tiny issues to fix elsewhere.

mixmix:
hey @christianbundy this branch is something @mmckegg and I were testing yesterday.
We sorely need scuttlebot 13 to be live, so quite keen to see this merged.

@mmckegg what's needed for this to be ready to go? please be explicit and write all the things at once :heart:

christianbundy:
Happy to hear this may be useful! Just to make sure it's documented, there are two regressions in this branch:
 
- **tagging:** the new scuttle-tag has a different API, I'm going to roll back now and let that work continue in #887 
- **DHT**: this just uses the default Scuttlebot gossip plugin, so I'd wager that all DHT-related functionality is broken. I've been meaning to look into resolving [this issue](https://gitlab.com/staltz/sbot-gossip/issues/1) but haven't found the time.

index.js
christianbundy: @mixmix I see that we're doing it [differently](https://github.com/ssbc/patchbay/blob/master/config.js) in Patchbay, so this section may require a revamp. I've only tested this on Linux where the unix socket works great.

christianbundy:
It looks like the sbot-gossip repo is just a small diff from Scuttlebot at [`plugins/gossip/index.js@41f303b`](https://github.com/ssbc/scuttlebot/blob/41f303be006b022da4a4fa3c08c365f548984a3a/plugins/gossip/index.js), which seems like it would be super simple to keep up-to-date if we just used a branch instead of a new repo.

@staltz Would it make sense to branch off of Scuttlebot with this change, use the DHT branch rather than the NPM module, and replace instances of `require('sbot-gossip')` with something like `require('scuttlebot/plugins/gossip')`? I think most of the hairiness with the sbot@13 upgrade comes from the fact that this is a repo rather than a branch, so I think that would make future maintenance much simpler.

mixmix:
To be clear @christianbundy , the main thing I want from this branch is sbot 13 ... and I think we'll add `ssb-about@2` which we extracted from patchwork's refactor yesterday. 

@mmckegg is fine with putting the DHT on hold / disabling it and revisiting it in another PR another time so that we can move fwd

christianbundy:
Oh, cool. In that case. I think this branch should be running fine. DHT won't work, but otherwise let me know if you see any funky behavior.

staltz:
> Would it make sense to branch off of Scuttlebot with this change, use the DHT branch rather than the NPM module, and replace instances of require('sbot-gossip') with something like require('scuttlebot/plugins/gossip')?

@christianbundy Part of my motivation with that package is to begin gradually migrating away from scuttlebot built-in plugins, and using more one-plugin-per-package. I really dislike the fact that scuttlebot comes with many (some complicated!) plugins out of the box, makes scuttlebot maintenance difficulty, and piecemeal upgrading difficult.

christianbundy:
I think that's a good idea! To be explicit, I completely support extracting plugins out of the scuttlebot module, my only preference is that we keep the Git history if possible. For example, I'd love [something like this](https://github.com/ssbc/scuttlebot/tree/scuttlebot-plugin-gossip) where we don't lose our history or the ability to `git merge`.

If you propose extracting plugins out of Scuttlebot I'd support that proposal, but that's probably a discussion for another issue. :heart:

mixmix:
`git-ssb-web` fails to start because it doesn't know how to connect to sbot any more:
https://git-ssb.celehner.com/%25q5d5Du%2B9WkaSdjc8aJPZm%2BjMrqgo0tmfR%2BRcX5ZZ6H4%3D.sha256/blob/3852c8c0341ee5a8a6170245a58ef56c3e574058/server.js#L19

I think ssb-config needs to set a default connection? Not sure

mmckegg:
@christianbundy I just merged in the latest master and removed the dht invite buttons (for now).

But I am seeing one interesting issue: When you first switch to this version of patchwork from the latest release, it reindexes, but never loads contacts. So you can't see that you are following anyone and a warning message is permanently stuck on the public view until you restart.

Otherwise than that, I think this is ready to merge.

mmckegg:
I'm gonna merge this and then add the following bug as another issue.

mixmix:
Yusssss. Im so pumped for this patchbay friendly patchwork release

index.js
christianbundy: Note to future travelers: this exposes your private key to the terminal output. It's harmless by itself, but if you copy and paste your terminal output from this commit to GitHub then you may expose your private key.

This bug has been fixed in `master`, but it's still in the Git history (obviously).

PR #902, opened 2018-12-13
namespace flume indexes

dangerousbeans:
if this goes in
https://github.com/ssbc/scuttlebot/pull/594

so we don't have to rebuild jazz all the time when app switching

### DIFF
--- a/index.js
+++ b/index.js
@@ -167,7 +167,8 @@ function setupContext (appName, opts, cb) {
     friends: {
       dunbar: 150,
       hops: 2 // down from 3
-    }
+    },
+    db_prefix: "patchwork",
     // connections: { // to support DHT invites
     //   incoming: {
     //     dht: [{ scope: 'public', transform: 'shs', port: 8423 }]


mmckegg:
@dangerousbeans  Hey, cool! 

I'm open to something like this, though it kind of misses the advantage of having shared indexes at all (which opens up another discussion). 

What would help for a start is to remove the unused indexes out of patchwork. Only using `ssb-backlinks` + the core sbot indexes (there is a reference to ssb-query right now, but this can easily be replaced).

mmckegg:
Oh, maybe a better way of handling this @dangerousbeans is to prefix the version of the index to the index name instead of a custom namespace. I think @dominictarr might have already suggested this.

mmckegg:
I'm going to close this as (for now) the solution is to use the same versions of indexes in patchwork/patchbay, but in the future we'll have `lessbot` and `followbot` to wrap this problem up. The final result will be similar to what this PR does, except no need for namespacing because all of an apps indexes will be in its own `.ssb` folder.

PR #898, opened 2018-12-06
Update har-validator

christianbundy:
har-validator@5.1.2 was unpublished, so this commit manually updates itto 5.1.3 to resolve an installation issue with the missing dependency.

Resolves #897 

cc: @Aversiste can you confirm that this resolves the issue for you?

PR #904, opened 2018-12-17
Update har-validator

neftaly:
I was unable to clone & npm install following a reformat. Turns out har-validator was reset back to 5.1.2 in package-lock, which has been unpublished. To reproduce this issue, git reset && rm node_modules && npm install.

This is a repeat of #898, to fix a reversion introduced in b947e38.

Can we make sure all developers reset their package-lock files and re-install NPM modules, to prevent another reversion from happening? Or better yet, enable branch protection and require PRs for all commits to master.

---
har-validator@5.1.2 was unpublished, so this commit manually updates it
to 5.1.3 to resolve an installation issue with the missing dependency.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3767,9 +3767,9 @@
       "dev": true
     },
     "har-validator": {
-      "version": "5.1.2",
-      "resolved": "https://registry.npmjs.org/har-validator/-/har-validator-5.1.2.tgz",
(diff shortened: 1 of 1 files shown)


neftaly:
I am not sure if NPM will attempt to install 5.1.3 first if it's cached locally (and it probably won't happen if 5.1.2 is cached), so this issue might only affect fresh clones/installs.

christianbundy:
LGTM, `npm ci` is passing now. Thanks!

neftaly:
Thanks, hopefully there isn't another regression before the week is out :P

PR #916, opened 2019-01-04
add key shortcut for submitting from compose

### LINKED ISSUE #813: Fix tab+enter to post comment behavior
My understanding is tab+enter generally means "post/enter/submit" in a lot of software, and my muscle memory certainly thinks so. Currently if you tab it jumps to the "attachment" option and enter opens the "pick file" dialogue. If it could jump to "reply" instead, that would be grand.

robjloranger:
A quick start, does not yet change anything about the confirmation dialog. That could always be in another PR.

Would it be of interest to have both `ctrl` and `shift` enabled? Something easy enough as:
```javascript
if (ev.keyCode === 13 && (ev.shiftKey || ev.ctrlKey)) {
..
}
```

Will close #813

### DIFF
--- a/modules/message/html/compose.js
+++ b/modules/message/html/compose.js
@@ -59,6 +59,12 @@ exports.create = function (api) {
         if (!files || !files.length) return
         attachFiles(files)
       },
+      'ev-keydown': ev => {
+        if (ev.key === 'Enter' && (ev.ctrlKey || ev.metaKey)) {
+          publish()
+          ev.preventDefault()
+        }
+      },
       disabled: publishing,
       placeholder
     })


mmckegg:
@robjloranger 

Sorry, I haven't had a chance to reply about [Shift vs Ctrl](https://github.com/ssbc/patchwork/issues/813#issuecomment-451291221).

I actually meant to say Ctrl+Enter in my response. However, you'll probably want to handle Cmd+Enter on macOS as this is the standard convention.

Could possibly get away with:

```js
if (ev.keyCode === 13 && (ev.metaKey || ev.ctrlKey)) {
..
}
```

**One more suggestion:** In modern browsers you can use [`ev.key`](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key) and [`ev.code`](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code) instead of `ev.keyCode` to get a string based name for the keypress. `ev.code` gives the actual hardware button pressed, whereas `ev.key` is the resolved key for the current keyboard layout.

`ev.key === 'Enter'`

robjloranger:
I can't test on osx or windows but it's still working with the change on linux.

mmckegg:
Yep, seems to work as expected on macOS, merged!

Issue #919, opened 2019-01-13
How to print debug info


Powersource:
There was some way to print more verbose error output, but I can't find that info again. Could someone add that info to e.g. https://github.com/ssbc/patchwork/blob/master/contributing.md ?

Powersource:
Found it! https://github.com/ssbc/patchwork/issues/673
`--logging.level info` or debug or something else.

PR #921, opened 2019-01-13
Add info about changing logging levels

Powersource:
Fixes https://github.com/ssbc/patchwork/issues/919


mmckegg:
Thanks @Powersource!

PR #934, opened 2019-02-05
Redact private key from startup output

### LINKED ISSUE #933: Should Patchwork `console.log` the keys on startup?
I didn't notice that https://github.com/ssbc/patchwork/commit/eea148725dd55d17c3f5cbcb7c13776ab9549645 actually changes some startup order:

## Before

1. Load basic config, no keys
2. Output current config
3. Add keys

## After

1. Load basic config and keys
2. Add OS-specific customization (depends on keys)
3. Output current config

## Result

The result is that right now the private key is being output to consoles, which could be a gigantic footgun for folks submitting bug requests who might accidentally expose their private key. Would anyone mind if I added a workaround to avoid printing the private keys?

christianbundy:
Previously we did this output *before* adding the keys to the config,
but recently there were changes made that required the keys to be added
to the config before the config was finished. This meant that the config
being output contained private keys, which seems like a major security
hazard for debugging.

This commit changes the code so that we make a copy of the full config,
set `redactedConfig.keys.private = null`, and then output the redacted
configuration safely without any secrets exposed.

---

Resolves #933

### DIFF
--- a/index.js
+++ b/index.js
@@ -196,7 +196,9 @@ function setupContext (appName, opts, cb) {
     ssbConfig.remote = `unix:${socketPath}:~noauth:${pubkey}`
   }
 
-  console.log(ssbConfig)
+  const redactedConfig = Object.assign({}, ssbConfig);
+  redactedConfig.keys.private = null
+  console.log(redactedConfig)
 
   if (opts.server === false) {
     cb && cb()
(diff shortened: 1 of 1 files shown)


christianbundy:
I'm going to quick-merge this because:

1. it's trivial
2. it's critical
3. I'd feel real bad if someone leaked their key because of me

Powersource:
Just noticed this was fixed, thanks! :tada:

PR #940, opened 2019-02-20
Add Ctrl/Cmd+L shortcut to go to search field.

black-puppydog:
I found myself hitting Ctrl+L quite often, since that's how all the browsers activate the URL bar.
I just looked at how #802 dealt with the Forward/Backward, and tried to adapt. If I understood correctly, the problem here was that I wanted the rendering process to change something in the DOM (i.e. give focus to the input) but was trying to set that up in the main process.

This is of course just an idea. Specifically, passing code as a string strikes me as smelly, but given that I speak close to zero JS, that's what I ended up with. So feel free to ignore this. :P

### DIFF
--- a/index.js
+++ b/index.js
@@ -57,8 +57,15 @@ electron.app.on('ready', () => {
     var menu = defaultMenu(electron.app, electron.shell)
 
     menu.splice(4, 0, {
-      label: 'History',
+      label: 'Navigation',
       submenu: [
+        {
+          label: 'Activate Search Field',
+          accelerator: 'CmdOrCtrl+L',
+          click: () => {
+            browserWindow.webContents.send('activateSearch')
+          }
+        },
         {
           label: 'Back',
           accelerator: 'CmdOrCtrl+[',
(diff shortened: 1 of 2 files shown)


black-puppydog:
just figured out what was making this less comfortable for me: I didn't select the search box content on Ctrl-L, so I had to Ctrl-A, Backspace every time. That's fixed now. I still find this dirty-but-useful. Any comments anyone?

mmckegg:
This seems like a very useful feature. Thanks for the PR @black-puppydog!

I would suggest that this was handled using `browserWindow.webContents.send('activateSearch')` (similar to how #802 implemented it in the end).

Then it would be possible to update [search.js](https://github.com/ssbc/patchwork/blob/master/modules/app/html/search.js) to listen for the event:

```
// insert after line 35
electron.ipcRenderer.on('activateSearch', () => {
  searchBox.focus()
  searchBox.select() // should handle selecting everything in the box
})
```

And then you can avoid the magic css selectors, and help make the code smell a bit fresher 💯

black-puppydog:
TIL. Thanks for showing a nicer way. It works for me. :)

PR #946, opened 2019-03-03
fix private blob URLs

ahdinosaur:
:smiley_cat:

private blob URLs were missing the unbox key, so the downloaded file was
encrypted gobbledygook. this change ensures the query string (with the
unbox key) is included in the URL.

---

i'm not sure if how i implemented this change is the best way, as in i'm
not sure if this solves one use case only to break another.


mmckegg:
Seems to work. Guess we'll have to keep an eye out for other weird side effects ⭕️

PR #949, opened 2019-03-09
replace gossip plugin with new rewrite

staltz:
I've been working on the gossip plugin refactor/rewrite. `ssb-legacy-conn` is essentially the same functionality as `ssb-gossip`, but with its internals reworked with a new stack of libraries (notably, `ssb-conn-db` migrates the file `gossip.json` to `conn.json`). I'm using `ssb-legacy-conn` also in Manyverse.

Try out this PR branch before merging. Everything should functionally be the same, the one difference is that `gossip.json` is *read* but never *updated*, only `conn.json` is updated from now onwards. I'll be updating `ssb-legacy-conn` gradually until its internals are very easy to work with. Then it'll be easy to create alternative plugins, e.g. `ssb-better-conn`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -42,36 +42,6 @@
         "prebuild-install": "^2.4.1"
       }
     },
-    "@staltz/sbot-gossip": {
-      "version": "1.1.6",
-      "resolved": "https://registry.npmjs.org/@staltz/sbot-gossip/-/sbot-gossip-1.1.6.tgz",
-      "integrity": "sha512-/WTNtBHfXwGB+v61ukIh6xB+Su9/jY+a2dTwRrDSU9BOcdFfpgA2cjrFR73hEXMxDGz6SDaHOb2njKkfy3b0Eg==",
-      "requires": {
-        "atomic-file": "0.0.1",
-        "deep-equal": "^1.0.1",
-        "ip": "^0.3.3",
-        "mdmanifest": "^1.0.4",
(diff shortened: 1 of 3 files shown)


christianbundy:
Please forgive my ignorance, I'm *super* excited about a refactored gossip plugin, but I'm wondering why this is being done as a module switcheroo rather than just a merge into ssb-gossip. Is this something that you'd be interested in merging into ssb-gossip, or would your preference be to maintain  your own module for it?

staltz:
@christianbundy I guess my reasoning is that ssb-legacy-conn is somewhere between ssb-gossip and ssb-conn, and this can more clearly be seen with the files (from `gossip.json` to `conn.json`). Also I tend to think that repositories are cheap, so I default to creating more instead of replacing. Also helps me move faster as I don't have to justify some tech decisions inside ssb-conn (such as TypeScript usage).

christianbundy:
Just an update, no action required: I'd really like to merge this but I've been prioritizing bug-squashing commits so that we can get a Patchwork release ASAP. My preference would be to merge this once we have a reasonably stable Patchwork release and I'm happy to take care of the merge conflicts when that time comes.

staltz:
Sure, no problem. The merge conflicts shouldn't be a big deal, it's just about using `ssb-legacy-conn@1.0.25` in package.json, then running npm install again so that it updates package-lock.json

PR #959, opened 2019-03-18
Backport npm updates

christianbundy:
- scuttlebot => ssb-server
- installed new ssb-server plugins
- upgraded patchcore, ssb-about, and ssb-ws
- added used modules that were missing from `package.json`:
  - lodash
  - run-parallel
- removed `sbot-local-widh-address-fix.js`

Issue #960, opened 2019-03-18
Patchwork 3.11.5 seems to be connecting to blocked pubs


mmckegg:
Not sure if this is a new bug (doesn't seem to exist in master). Most likely something to do with the backports. 

@christianbundy Can you check and see if this is fixed in #959 ?

christianbundy:
I just noticed this in my branch, it's not fixed. Not sure where it came from though...

![Screenshot from 2019-03-18 14-01-39](https://user-images.githubusercontent.com/537700/54563453-72ae9480-4986-11e9-82ff-ec4dac2cd2b5.png)

christianbundy:
I've downgraded from ssb-server@14 to ssb-server@13 and I think this is resolved -- not sure why it was happening, but it seems fine now.

christianbundy:
Fixed in #961

mmckegg:
@christianbundy  I think that maybe ssb-server just does this, not 100% sure though.

This is fixed in patchwork@master but I think that is because patchwork is in control of the peer table (as part of the ssb-friends rewrite), and it [excludes blocked peers](https://github.com/ssbc/patchwork/blob/master/sbot/index.js#L140) from the gossip table.

christianbundy:
Fixed in `master` and prereleases. :tada:

PR #961, opened 2019-03-18
Backport ssb-server@13 to Patchwork 3.11

christianbundy:
This is a more minimal take on #959, bringing Patchwork up-to-date with ssb-server@13 without trying to make the leap to the bright and shiny ssb-server@14. 

- scuttlebot@12 => ssb-server@13
- ssb-ws@2 => ssb-ws@6
- fixed missing/unused modules in `package.json`:
  - added lodash
  - added run-parallel
  - removed node-abi
- removed `sbot-local-widh-address-fix.js`

---

The goal here is to more closely match `master`, **not** a full upgrade to the latest and greatest.


christianbundy:
This is also affected by #960, I'm not sure why. :unamused:

waynew:
Does this mean the dht invites (should be able to) work again?

christianbundy:
I'm not sure, I don't remember what the blocker for that was. I thought that they only worked on ssb-server@12 or something but I don't have a clear memory.

christianbundy:
I'm thinking/hoping that we'll get a release from `master` soon. Closing this now because I think it was giving Matt "weird perf issues'.

PR #963, opened 2019-03-19
remove unused make-single-instance.js

nornagon:
i couldn't find anything that used it anyway 😅


christianbundy:
Nice find, thanks!

PR #964, opened 2019-03-19
upgrade to electron@4.1.0

nornagon:
Just the one breaking change afaict!

Electron 2 is based on Chrome 61, this brings Patchwork to Chrome 69.


nornagon:
I'm seeing ```error loading sodium bindings: No native build was found for runtime=electron abi=69 platform=darwin arch=x64
falling back to javascript version.```

when running this, but unsure if that's a real issue or if that's just me having failed to properly build sodium. still seems to work ok though!

mmckegg:
@nornagon Have you tried running `npm run rebuild`? That should rebuild the native deps with the new electron version abi.

---

I seem to recall that the holdup on updating to newer electron version was that there were no prebuilds of various dependencies (spellcheck, sodium, leveldb) for the new ABI. We need these prebuilds otherwise can't do builds for multiple platforms (and it makes `npm install` much slower and more difficult for folks).

So for this to be merged, we'd need to check if all of the native deps now have prebuilds, and if not (in the case of spellcheck - for which we do [manual prebuilds](https://github.com/ssbc/paulcbetts-spellchecker-prebuilt)) update the config so that they do get prebuilds.

christianbundy:
Maybe we could upgrade to Electron 3 instead? Upgrading to Electron 4 would break all of the Arch Linux packages as well.

nornagon:
Based on this list of native modules:

```
$ rg -l node-gyp node_modules -g package.json
node_modules/node-gyp-build/package.json
node_modules/leveldown/package.json
node_modules/sodium-native/package.json
node_modules/utp-native/package.json
node_modules/nan/package.json
node_modules/@paulcbetts/spellchecker/package.json
node_modules/@paulcbetts/cld/package.json
node_modules/keyboard-layout/package.json
```

- leveldown@5.0.0 uses napi: https://unpkg.com/leveldown@5.0.0/prebuilds/darwin-x64/
- sodium-native@2.2.6 has prebuilds for electron-69 (i.e. 4.x) https://unpkg.com/sodium-native@2.2.6/prebuilds/darwin-x64/
- utp-native@2.1.3 uses napi: https://unpkg.com/utp-native@2.1.3/prebuilds/darwin-x64/
- @paulcbetts/spellchecker isn't prebuilt
- @paulcbetts/cld isn't prebuilt
- keyboard-layout isn't prebuilt

so it looks like the only upstream prebuilt binary that needs upgrading for 4.x is sodium-native.

It looks like patchwork currently depends on sodium-native via this chain:

```
ssb-patchwork@3.12.0-dev /Users/nornagon/Source/patchwork
└─┬ ssb-keys@7.1.3
  └─┬ chloride@2.2.10
    └── sodium-native@2.2.2
```

though chloride@2.2.10 actually depends on `sodium-native@^2.1.6`, so it looks like the package-lock.json was manually edited at some point.

I've opened https://github.com/dominictarr/chloride/pull/11 to get `chloride` to declare a newer dependency for sodium-native, once that's merged & released I'll do the same for `ssb-keys`.

In the mean time we could just force `sodium-native` to 2.3.0.

I'll look into bumping the prebuilds for the spellchecker, too.

nornagon:
ah, actually it looks like `sodium-native` was updated to 2.3.0 in bfe4c2f20

nornagon:
looks like the CI failures on node 4 and 5 are also present on master, & not caused by this PR.

package.json
```diff
@@ -47,6 +47,7 @@
     "moment-timezone": "^0.5.23",
     "mutant": "github:mmckegg/mutant#intersection-binding-viewport",
     "mutant-pull-reduce": "^1.1.0",
+    "node-abi": "^2.7.1",
```
christianbundy: Is this used somewhere?

christianbundy:
I'm running this locally and it's working great. Thanks for the contribution! My only hold-up is based on Matt's above comment:

> We need these prebuilds otherwise can't do builds for multiple platforms

I'm going to do some work on our build pipeline, so *once we have that* then I'll be happy to merge this (assuming that it passes the build pipeline). Super excited to get this merged!

nornagon:
Fantastic, let me know if there's anything I can do to help!

package.json
nornagon: This is depended on transitively, I added this minimum dependency to force the upgrade. I suppose it could be removed now that package-lock.json has updated with the higher version. (The earlier version doesn't support Electron 4.)

christianbundy:
Taking a look at this again, are you seeing this in the console output? If so, I think we'll need to fix that so that we're not using JavaScript cryptography (because it's slow).

> error loading sodium bindings: No native build was found for runtime=electron abi=69 platform=linux arch=x64

christianbundy:
Merged! Thanks a bunch, this is working great and you've paved the way for future Electron updates. For future reference, merging only required that I replaced references to the old ABI (57) to the new one (69) in `package.json`. Building correctly was handled by electron-builder, which solved the "error loading sodium bindings" thing above. I'll plan on making a new alpha prerelease early next week.

:shipit:

nornagon:
Fantastic! Electron 5 was just released: https://electronjs.org/blog/electron-5-0 so I'll have a go at updating to that soon.

christianbundy:
Yep, that's what I was working on when I noticed [this issue](https://github.com/electron/electron/issues/17972), and for some reason `npm start` wasn't opening a window after the chown/cmod was finished. Didn't debug, too excited about Electron 4 (:tada:)!

Feel free to push to a topic branch (maybe `electron-5`?) in this repo, that should get Travis to auto-test your branch and makes it *slightly* more convenient for others to pull for review.

gergelypolonkai:
As far as i know this PR obsoletes my older PR, #922. I’ll give it a go as soon as i can.

PR #971, opened 2019-03-25
Handle magnet URLs

dangerousbeans:


lib/catch-links.js
```diff
@@ -18,6 +23,14 @@ module.exports = function (root, cb) {
...
+      // Check for magnet link
+      if(magnet_pattern.test(href))
+      {
+        // throw to OS (safe?)
+        shell.openExternal(href)
+        return
+      }
+
       try {
         var url = new URL(href)
```
mmckegg: Whoops, github nuked my other review note.

This is already parsing magnet urls, but they are getting routed as 'internal' urls. To change this, you can update the check to something like this:

```js
if (isUrl && (url.host || url.protocol === 'magnet:')) {
```

Or if you want to keep the regexp check:

```js
if (isUrl && (url.host || (url.protocol === 'magnet:' && magnet_pattern.test(href)))) {
```

dangerousbeans: oh ya that makes sense! pushed a simpler version

dangerousbeans:
Hm this borked it, sometimes url doesn't exist

dangerousbeans:
okay now it's beautiful

lib/catch-links.js
```diff
@@ -25,7 +27,7 @@ module.exports = function (root, cb) {
         isUrl = false
       }
 
-      if (isUrl && url.host) {
+      if (isUrl && url.host || isUrl && url.protocol === 'magnet:') {
```
mmckegg: @dangerousbeans 

or `isUrl && (url.host || url.protocol === 'magnet:')`

mmckegg:
@dangerousbeans another thing worth doing here is to update the external link preview (when hovering) to show magnet urls: see [modules/app/link-preview.js#L50](https://github.com/ssbc/patchwork/blob/master/modules/app/link-preview.js#L50)

Change it to something like this:

```js
        } else if (href.includes('://') || href.startsWith('magnet:')) {
```

Chery46:
I think you should use an
`<addr>` element here instead.

Chery46:
I think you should use an
`<addr>` element here instead.

christianbundy:
This looks great to me, I've confirmed that it works and I've merged the small logic refactor that Matt requested above. Thanks for the contribution @dangerousbeans!

PR #974, opened 2019-03-27
Use read-only mode while indexing

### LINKED ISSUE #906: add optimistic updating back in for likes, follows, subscribes
Or at least show in the interface that we are waiting for the backend to finish

When sbot is running slow, or during initial onboarding, clicking the various user action buttons does not update the user interface immediately.

This is a regression caused by #858. 

Ideally we'd fix it with some kind of holding tank solution, but it might just have to be a UI thing for now.

christianbundy:
This adds some simple CSS that:

- :hourglass_flowing_sand: Sets `cursor: progress` so the user knows that indexes are running in the background.
- :hand: Sets `cursor: wait` if the user selects an element that will publish a message.
- :x: Sets `cursor: not-allowed` if the user tries to activate an element that will publish a message.
- :ghost:  Sets `opacity: 0.5` on these elements to mimic [`disabled`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Input#disabled) links
- :exclamation: Outputs an error on any attempt to publish while the indexes are still building.

Resolves #906 

Screenshot from 2019-03-27 10-40-57

Screenshot from 2019-03-27 10-41-35


christianbundy:
Related: I've resolved this problem another way in 05bcda3c0050242bcc340f188e1159520fe1756a by modifying `sbot.publish` to ensure that the indexes are all up-to-date. It works great at preventing publishing, but I can't figure out how to check the indexes while they aren't changing (see https://github.com/mmckegg/mutant/issues/27).

I like the pattern better but it doesn't let you publish because when the indexes are up-to-date they aren't calling the observable. :/

Would love some help on this, I'm super hyped to get a new Patchwork release.

christianbundy:
Thanks, merging now! One down, only a few more to go.

Issue #952, opened 2019-03-13
Dark Crystal Integration - Phase 1


kyphae:
Shard and share your `.ssb/secret` and your `gossip.json` directly in Patchwork. As discussed with @mmckegg, this will be a 3 phase integration, allowing users to:

- (a) backup their identity with their peers
- (b) reclaim their old identity from their custodians using a newly generated identity after it has been lost (lost / broken laptop).
- (c) restart Patchwork as the new identity, which automates reintegrating the recovered identity. Gives the user the option to remove or backup the identity in use.

# Phase 1

**Backup your identity with your peers**

### Backend
* [x] Extend the `shard` schema to accept a blob reference to `gossip.json` file in [`ssb-dark-crystal-schema`](https://github.com/blockades/ssb-dark-crystal-schema)
* [x] Extend the `shard` publish action params in [`scuttle-dark-crystal`](https://github.com/blockades/scuttle-dark-crystal) to accept a blob reference
* [x] Ensure the file is encrypted when creating a dark crystal

### UI
* [x] Draw some wireframes and run them by @mmckegg (upload here...)
* [x] CSS! Consult Matt to make sure it fits with Patchwork's overall style

#### Required functionality
* [x] Add a `Back up` button to the `/settings` page which:
* [x] Opens a modal containing a form that takes 2 inputs: a list of custodians and a quorum
* [x] Form submission performs the following
  * [x] Load the entire `.ssb/secret` and stringify the JSON - this is the secret
  * [x] Load, encrypts and generate a blob ID for the `gossip.json` file
  * [x] Uses `scuttle-dark-crystal` to publish the message with `SSB Identity` as the `name`.
* [x] Add details about any existing backup so users can see that they've successfully backed up their identity - current custodians and the quorum required.

### Backlinks:
Phase 2 - https://github.com/ssbc/patchwork/issues/953
Phase 3 - https://github.com/ssbc/patchwork/issues/954

ameba23:
@KGibb8 

> Extend the root or the ritual schema to accept a blob reference to gossip.json file in ssb-dark-crystal-schema

This excites me, as allowing bigger secrets by using blobs is something we've often talked about but never managed to make happen.  and i think it would be a great feature to have.

But i would argue that since the purpose of this design is account recovery, there's not much point in referencing the blob in the root or ritual message, since they can only be seen by the owner of the secret.

i would propose the blob reference should either be included in each shard message, or be included in the secret itself.  

[...]

PR #976, opened 2019-03-28
[WIP] Dark crystal

KGibb8:
See https://github.com/ssbc/patchwork/issues/952


ameba23:
@KGibb8 in terms of changing `custody.js` to use the query [forwardRequest.pull.forOwnShards](https://github.com/blockades/scuttle-dark-crystal/blob/forward-request/forward-request/pull/for-own-shards.js), i think we might have to make some changes to the data structure you were putting together (`store` and `records`).  i think it would maybe be easier if they were indexed by 'secret owner' rather than the id of the shard message. but i wanted to check with you before changing it.  see commented out code in the last commit.

KGibb8:
We can't do it like that if we want to display all shards you've been asked to be the custodian of. 

First we have to display a list of shards you hold, that contain an attachment (relevant to Patchwork), then we need to search for forwardRequests that match up to those specific shards (we cross-reference using secretOwner on the request field, with the shard author). 

Otherwise this view: https://user-images.githubusercontent.com/17786217/54701493-05893380-4b2d-11e9-9b5d-e5158f50a97c.jpg  won't work

PR #980, opened 2019-04-03
Use ddg instead of google for context menu search

Powersource:
Boom, tested and works.


Powersource:
Ok soo travis fails but it's only on old js versions and on code I didn't write :P

mmckegg:
Awesome, seems to work! Merging 😸

PR #984, opened 2019-04-07
Only show search in the context menu when relevant

Powersource:
Had a small bug in https://github.com/ssbc/patchwork/pull/980 hehe


christianbundy:
@Powersource What was the bug fixed here?

christianbundy:
Nevermind, it looks like previously there was an issue where "search with ddg" was showing up *everywhere*, so this PR fixes it so that it only shows up where previously there was a google option. Thanks for the contrib!

Powersource:
Oh yeah I'll mention *what* bug I'm fixing next time 😅 Thanks for merging!

PR #989, opened 2019-04-13
Change rendering of posts with authors blocked by the thread author

Powersource:
Initial proposal and discussion `%IZHgIL3MuP2rfxLncnisRdnIsK7H0Wc+SDviVGDuGzg=.sha256`

~~Still WIP, but some slight code review already probably wouldn't hurt :)~~ Review away!

TODO:
* [x] I18n, how does it even work lol
* [x] Don't just replace the post text, render it in the same way that missing messages are. Would probably fix most of the uglyness (including the weird show less/more bug.


Powersource:
Ready for review now! There are a few things I suspect maybe should be changed (maybe the css?), but I don't know where the bar goes for "it's ugly but it works, ship it!" :smile:

Powersource:
Oh yeah I'll post the image here as well

image

christianbundy:
Thanks for the screenshot! I think this is great behavior, but I'd *love* for this to reuse the visual style of references and forked discussions. Do you think that would be a simple change?

christianbundy:
I think this might still have a few visual glitches going on, as these messages are missing their borders:

Screenshot from 2019-04-23 13-33-05

Screenshot from 2019-04-23 13-34-17

I mentioned this above, but I think the most elegant solution here would be to show these blocked replies as a "forked discussion" -- I'd be curious to know what you think.

Powersource:
That would kind of make sense (displaying the same way as forks/mentions). Just wondering what to do when the text gets too long for one line. Mentions/forks solve it by truncating the message preview. At least we could shorten the text to "Post by Alice hidden because they're blocked by the thread author Bob" and have the whole thing be a link to the fork. It seems this would fit with most usernames, maybe it doesn't hurt that it occasionally wraps to an additional line?

christianbundy:
> Mentions/forks solve it by truncating the message preview.

Could we copy this behavior? Ideally we could reuse the same code so that we don't have to implement it in two places.

Powersource:
There we go, was easier than I thought
image
and doesn't feel as confusing as I'd imagined when grouped with forks and references
image
and plenty of space for long PREMIUM usernames (even looks pretty good when wrapping).

Pushing in a sec.

Powersource:
Problem with truncating is that it would remove context for why the message is hidden.

Powersource:
Assuming the tests pass now, this is ready for review again :)

christianbundy:
Sweet! I'm still seeing some differences from the visual style, I'll highlight it below:

## Fork/reference

> **[$name](#) forked this discussion:** $preview

## Blocked

> This post by [$blockee](#) is hidden because they are blocked by the thread author [$blocker](#).

## Proposed

Maybe we could do something like:

> **[$blockee](#) replied but is blocked by [$blocker](#):** $preview

I really appreciate your work on this, I know everyone is excited to have this resolved.

Powersource:
Ah ok you mean like *that*. I'll look into it.

Powersource:
image
Pushing in a sec.

lib/filter-blocked.js
```diff
@@ -1,25 +1,54 @@
...
-      if (err) return cb(err)
-      cb(null, !blocking)
-    })
-  }, 10)
+        // Important that we start with 0 (us), we're prioritized above thread
+        // authors
+        for (let i = 0; i < checks.length; i++) {
+          if (res[i]) {
+            // I shouldn't be modifying the original msg in a (para)map I guess
+            // but :shrug: (? or will I break stuff)
```
christianbundy: You can use `msg.value.meta` for this, see also: https://github.com/ssbc/ssb-db/pull/228#issuecomment-435438060

I'd recommend `_.get()` and `_.set()`, for example:

```js
const _ = require('lodash')

_.set(msg, 'value.meta.isBlocked', true)
```

This way you can also avoid this sort of thing:

```js
msg.value.meta && msg.value.meta.blockedBy && msg.value.meta.blockedBy.id

// vs

_.get(msg, 'value.meta.blockedBy.id')
```
```

File: lib/filter-blocked.js [hunk 2]
```diff
@@ -1,25 +1,54 @@
...
+            return cb(null, msg)
+          }
+        }
+        cb(null, msg)
+      })
+    }),
+    pullFilterAsync((msg, cb) => {
+      // Only completely remove the msg if we are the blocker
+      cb(null, !(msg.isBlocked && msg.blockedBy.role === 'me'))
+    }, 10)
```
christianbundy: Could we use a synchronous [filter-not](https://pull-stream.github.io/#filter-not) here? I think it might look something like:

```js
    pullFilterNot((msg) =>
      // Only completely remove the msg if we are the blocker
      msg.isBlocked && msg.blockedBy.role === 'me'
    )
```

modules/message/html/render/post.js
```diff
@@ -19,13 +20,16 @@ exports.gives = nest({
...
+
   return nest('message.html', {
     canRender: isRenderable,
     render: function (msg, opts) {
       if (!isRenderable(msg)) return
+
       var element = api.message.html.layout(msg, extend({
         title: messageTitle(msg),
-        content: msg.isBlocked ? 'Content of a blocked user' : messageContent(msg),
+        content: msg.blockedBy && msg.blockedBy.role === 'me' ? i18n('Content of a blocked user') : messageContent(msg),
```
christianbundy: Nice i18n!

christianbundy: Line 32 is pretty complex, maybe we could simplify? For example:
```js
const isBlocked = msg.blockedBy && msg.blockedBy.role === 'me'

var element = api.message.html.layout(msg, extend({
  title: messageTitle(msg),
  content: isBlocked ? i18n('Content of a blocked user') : messageContent(msg),
  layout: 'default'
}, opts))

File: modules/page/html/render/message.js [hunk 4]
```diff
@@ -229,15 +249,24 @@ exports.create = function (api) {
...
       var msg = { key: id, value }
-      if (isBlog(msg)) {
-        Blog(api.sbot.obs.connection).async.get(msg, (err, result) => {
-          if (err) return cb(err)
-          msg.body = result.body
-          cb(null, msg)
+
+      var me = api.keys.sync.id()
+      onceTrue(api.sbot.obs.connection, sbot => {
+        sbot.patchwork.contacts.isBlocking({ source: me, dest: value.author }, (err, blocking) => {
```
christianbundy: What does this do? I'd love some comments here explaining the behavior that this creates.

modules/page/html/render/message.js
```diff
@@ -229,15 +249,24 @@ exports.create = function (api) {
...
+          if (err) throw err
```
christianbundy: We usually do something like:

```js
if (err) return cb(err)
```

Do we need to throw here?

modules/page/html/render/message.js
```diff
@@ -229,15 +249,24 @@ exports.create = function (api) {
...
+          if (blocking) {
+            cb(null, null)
```
christianbundy: Would love a comment / explanation here as well (assuming other comments don't explain it).

lib/filter-blocked.js
```diff
@@ -1,25 +1,54 @@
...
-  return pullFilterAsync((msg, cb) => {
-    var checks = []
-    blockSources.forEach(source => {
-      if (msg.value) {
-        checks.push({ source, dest: msg.value.author })
-      }
-      if (checkRoot && msg.root && msg.root.value) {
-        checks.push({ source, dest: msg.root.value.author })
+  // Maybe hacky but it seems true in practice
+  const me = blockSources[0]
```
christianbundy: Could we use `api.keys.sync.id()` or something similar here instead?

Powersource:
Thanks for the review! Gonna dump some quick comments before going to bed and forgetting stuff.

locales/en.json
```diff
@@ -303,5 +303,11 @@
...
-	"Error": "Error"
-}
+	"Hidden message": "Hidden message",
+	"This post by ": "This post by ",
+	" is hidden because they are blocked by the thread author ": " is hidden because they are blocked by the thread author ",
+	"Click here": "Click here",
+	" to view the post in a fork of this thread.": " to view the post in a fork of this thread.",
+	"Error": "Error",
+	" replied but is blocked by ": " replied but is blocked by "
+}
```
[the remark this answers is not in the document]
Powersource: Assumed i18n did that for me, will prune.

modules/page/html/render/message.js
Powersource: I think I saw standard complain about this and I just put a throw in there by habit, still not very used to callbacks :smile:

modules/page/html/render/message.js
Powersource: I think I looked up how to get the page to render the "Can't display message" page, and this was how to do it. Will look into and add a comment.

modules/page/html/render/message.js
Powersource: I think this whole blob of code was me thinking a bug (direct links to blocked people's posts still rendered) was my fault and trying to fix it. Turned out it was in all along, I think, but now it's fixed anyway hehe.

If you're asking about the whole onceTrue thing that's a piece of code I found somewhere else, I needed an sbot connection (still not clear on why haha) to get the isBlocking function.

lib/filter-blocked.js
Powersource: I thought about doing that, but there's no `api` object access in this file and I don't know how to get that.

lib/filter-blocked.js
```diff
@@ -1,25 +1,54 @@
...
-      cb(null, !blocking)
-    })
-  }, 10)
+        // Important that we start with 0 (us), we're prioritized above thread
+        // authors
+        for (let i = 0; i < checks.length; i++) {
+          if (res[i]) {
+            // I shouldn't be modifying the original msg in a (para)map I guess
+            // but :shrug: (? or will I break stuff)
+            msg.isBlocked = true
```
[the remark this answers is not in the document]
Powersource: I think I did this because some things use the isBlocked variable. But maybe not anymore after these changes.

lib/filter-blocked.js
```diff
@@ -1,25 +1,54 @@
...
+            msg.blockedBy = {
+              id: checks[i].source,
+              role: checks[i].source === me ? 'me' : 'threadAuthor'
+            }
```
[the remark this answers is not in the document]
Powersource: I tried to clarify that on line 32, our own blocks have higher priority. Since they remove the post completely we don't need to worry about others having blocked them as well.

modules/page/html/render/message.js
Powersource: Satisfying explanation? :P Want any of it as comments in the code?

lib/filter-blocked.js
Powersource: Got a good solution here?

lib/filter-blocked.js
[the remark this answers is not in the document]
Powersource: Resolved?

Powersource:
Fixed/replied to all the review comments now :)

lib/filter-blocked.js
christianbundy: Nope, I think this is probably as good as we'll get -- the dichotomy between `lib` and `modules` kills me, so much code that can't be shared!

lib/filter-blocked.js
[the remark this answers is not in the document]
christianbundy: Yep, that makes sense. I think I might have been confused about what `checks` means, but now I get that it means "blocks" I think?

modules/page/html/render/message.js
christianbundy: Nope, I think this bit makes sense to me now (although comments are always appreciated!). :+1:

modules/page/html/render/message.js
christianbundy: It says this conversation was resolved but I still think that we want to callback the error. Usually callbacks take `cb(err, val)` where `err` is either `null` or an error object and `val` is the value that you'd otherwise `return` if it wasn't an async callback function.

modules/page/html/render/message.js
christianbundy: Looks like this was marked "resolved" but I'd still love a comment here.

lib/filter-blocked.js
christianbundy: Looks like this convo was marked resolved as well, does this need to be async though?

lib/filter-blocked.js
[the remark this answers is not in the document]
christianbundy: Sweet, I'd love if we could remove any variables we aren't using.

modules/page/html/render/message.js
Powersource: I think you're looking at old code here, it's currently `if (err) return cb(err)`

lib/filter-blocked.js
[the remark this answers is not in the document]
Powersource: Yeah checks as in 'checked if they were blocked'. Renamed it now tho.

lib/filter-blocked.js
Powersource: Looking at old code :P

Powersource:
Aaand done again :)

christianbundy:
Both the code and the interface look lovely and I'm so appreciative that you took this work on. Thanks for all of the hours you spent on this PR, I think this resolves a major pain point in Patchwork and I'm super excited to click the big green button and :shipit:.

:tada: :tada: :tada: :tada: :tada: :tada:

PR #990, opened 2019-04-14
[WIP] Add &-mentioning

Powersource:
Been missing this! It's not super pretty but it works, feedback on how to improve it would be appreciated :)

Todo:
* [ ] Add a ~~spinner~~[throbber](https://en.wikipedia.org/wiki/Throbber) for the loading images (if we're gonna have them in the list at all, see below).
* [ ] Maybe not render the missing images to begin with, but append them to the suggestion list when we have them.
* [ ] Fix the " "file not found" image is breaking out of its container"-bug. Maybe don't add broken images (can we detect that?) but this is at least relevant for non-image blobs (iirc).
* [ ] Make the suggestion box have a max width. Maybe just truncate the strings (is there maybe fancy css for this?).
* [x] Image captions are displayed in a monospace font instead of Patchwork default


christianbundy:
I love the idea behind this! I don't have time this morning to check it out, but I can take some time to do that before the week is over. I'm curious about this bit:

> It's not super pretty [...]

I'm not familiar with ssb-meme, although based on "&-mentioning" I'm guessing this is the blob searching from Patchbay?

Powersource:
Yeah it's like the blob searching from patchbay, you type '&' and then words that search for blobs, the blobs show up in a list like with emoji autocompletion.

Ugly things include  
* the suggestion box becoming super wide when the image text is super long (thx dan :P ). Should be easy to fix.
* when there's a blob that you don't have yet, there's just an empty square and you don't know if things are broken or if the image is just loading. Not sure how (design- nor implementation-wise) to fix this.

gergelypolonkai:
> * when there's a blob that you don't have yet, there's just an empty square and you don't know if things are broken or if the image is just loading. Not sure how (design- nor implementation-wise) to fix this.

Design-wise i’d vote for a spinner-like thing, although an animated thing, especially if there are many blobs in the list, might hog a lot of CPU power.

christianbundy:
I'm happy to merge ssb-suggest, but I'd really like to make sure we nail the design and ergonomics of ssb-meme before adding it as a new feature.

> when there's a blob that you don't have yet, there's just an empty square and you don't know if things are broken or if the image is just loading. Not sure how (design- nor implementation-wise) to fix this.

My intuition is that ssb-meme should probably check [`multiblob.has()`](https://github.com/ssbc/multiblob#blobshashash-cb) before including it in the list, but that sounds like it would be a downstream change.

Powersource:
I like that it includes blobs you don't have yet, lets you find some pretty exotic images. How about sorting it so that the ones we have are at the top, and the ones we don't have are at the bottom with a spinner?

christianbundy:
I think I might be misunderstanding, how can you use blobs that you don't have? In my mind you'd be able to see the caption, but since you don't have the blob I'm not sure how or why you'd attach an unknown blob in your post.

I think there are at least four bugs right now that are important to resolve:

- thumbnails are empty while loading, a loading indicator would be great
  - are we loading the full image? if possible I'd love to use an actual image thumbnail here
- the "file not found" image is breaking out of its container
- image captions are displayed in a monospace font instead of Patchwork default
- as you mentioned above, long descriptions make the suggestion box *huge*

Screenshot from 2019-04-18 08-17-18

Powersource:
> how can you use blobs that you don't have

There are 2 ways.  
1. is that once it's in the suggestion list and patchwork tries to render it, sbot tries to fetch it, so it *does* render, just after a little bit.  
2. is that you don't wait for it to render, you just hope it's a nice image and click send. Dan does this from time to time I believe and it's nice, but sometimes it can end poorly `%yLgZ+rrAHLwSS0vQUqzBETzveARWByCA5zn6nho0rtE=.sha256` :sweat_smile: 

Case 2 is maybe not super important to support, but I like 1. Maybe a compromise would be to only show blobs we have, but load missing ones in the background and append them over time? Idk how I would implement the reactive list though (maybe mostly because I'm bad at reactive stuff).

> are we loading the full image? if possible I'd love to use an actual image thumbnail here

Yeah the full image. Would generating a thumbnail be to improve performance (would it?)?

christianbundy:
@Powersource  Should this remain open?

Powersource:
Yes pls

Jacob

christianbundy:
Hey @Powersource, just looping back around on this because it's still open -- is this still a branch you're taking care of? I'm trying to make sure all of the SSBC/Patchwork pull requests are being taken care of.

cinnamon-bun:
Hello I just noticed this thread!

If this is still being worked on, can we make it include a caption/link to the original post where the blob was found?  I believe the social impact will be overall better if images come with attribution by default.  Otherwise artists and people posting personal photos will be sad when their images end up everywhere with no link back to the person who made them.

Powersource:
@christianbundy no not taking care of this atm. *I'd like to*, but it would be in march at the earliest, and then preferrably with funding to make sure it happens.

@cinnamon-bun I think andré has plans to develop a module that lets you find the first appearance of a blob, maybe we could wait with this feature until he's done that? Would make it a lot easier to implement.

christianbundy:
Sorry to change my mind, but I realized that since all of these PRs are outdated I've been ignoring the PR count and missed some new PRs. I'm going to close this for now, please re-open if you have the time / interest / energy to work on this again! :heart:

PR #995, opened 2019-04-17
Disable ssb-ebt (broken)

### LINKED ISSUE #977: syncing gets stuck and requires restart during onboarding
This is due to a [bug in ebt](https://github.com/ssbc/ssb-ebt/issues/22) (that patchwork is now using). We either need to wait until the bug in ebt is fixed before we can release, or find some temporary hack (e.g. reconnect to pub if ebt progress stalls for an extended period of time).

@christianbundy Another release blocker!

christianbundy:
This reverts commit f03d884d4db127b92665d3631474bf1180e7beb8.

Resolves #977, although if this is merged I'll likely add a "re-enable ssb-ebt" issue to track that enhancement.

### DIFF
--- a/sbot/index.js
+++ b/sbot/index.js
@@ -185,7 +185,7 @@ exports.init = function (ssb, config) {
       for (var from in data) {
         for (var to in data[from]) {
           var value = data[from][to]
-          ssb.ebt.block(from, to, value === false)
+          // ssb.ebt.block(from, to, value === false)
         }
       }
     })
--- a/server-process.js
+++ b/server-process.js
@@ -39,7 +39,7 @@ var createSbot = require('ssb-server')
   .use(require('ssb-search'))
   .use(require('ssb-ws'))
   .use(require('ssb-tags'))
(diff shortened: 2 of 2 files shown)


mixmix:
@dominictarr christian and I talked about this a lot. We want `ssb-ebt` really bad and love it and would love to have a call about how we can get it back in.

For the moment the problem is that it's breaking the stability of the current release and we need to get a stable release out with some of those security patches

dominictarr:
okay I was recently trying to add a feature to ebt... and got stuck because everything was too messy, have been working on refactoring and documenting...

mixmix:
Good to know. We're not wanting to lump all the work on you. Would you like
to do a call and carve it up?
Ebt is so awesome!

dominictarr:
@mixmix well, since you appear to be volunteering, my goal is to add _fork proofs_, but then I got stuck because there are a bunch of quirks or weird JGIWs (just get it working). So i'm going through, figuring out what I can document, and also simplifying things that don't break the tests. Do you see anything you can help with?

PR #1017, opened 2019-05-02
Fix profile sheet edit issue with undefined id

### LINKED ISSUE #1015: patchwork prompting to setup new profile on every load in latest master
Screen Shot 2019-05-02 at 10 07 28 AM

This dialog is popping up on startup, but otherwise it seems to be fine.

christianbundy:
Resolves #1015

### DIFF
--- a/main-window.js
+++ b/main-window.js
@@ -61,7 +61,12 @@ module.exports = function (config) {
 
   // prompt to setup profile on first use
   onceTrue(api.sbot.obs.connection, (sbot) => {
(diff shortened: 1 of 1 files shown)


christianbundy:
Thanks, I'll merge now.

I'm also confused at why that API changed between sbot 12 and 13. Weird. I looked through the source and didn't see any other usage, so I'm hoping there aren't any other breakages.

(btw, force-pushed to remove my dependency change commit)

Issue #1009, opened 2019-04-29
new alpha installer "nuked" the install folder - 3.12.0-alpha.3


ThunderFD:
wanted to see if the new alpha installer works.... well big mistake :')

OS: Win10 64bit

it somehow nuked the permission to edit the Programs\ssb-patchwork folder (I think I somehow managed to delete the contents though)

after running 3.12.0 alpha 3 both installers don't work anymore  (and 3.12.0 broke during install)
![image](https://user-images.githubusercontent.com/17814079/56924528-68c39980-6acd-11e9-99ac-27de70a26a5d.png)

I don't seem to have any permissions to modify the patchwork folder at all
![image](https://user-images.githubusercontent.com/17814079/56924649-b8a26080-6acd-11e9-93e1-ef60e0c16b3b.png)

even as admin, trying uncheck read only mode 
![image](https://user-images.githubusercontent.com/17814079/56924800-15058000-6ace-11e9-9400-367834faa6ae.png)

![image](https://user-images.githubusercontent.com/17814079/56924848-36666c00-6ace-11e9-9410-89f2cd537100.png)
![image](https://user-images.githubusercontent.com/17814079/56924871-41210100-6ace-11e9-8182-fbb320e1ab98.png)

no idea how I can modify the folder now..
any attempts to change anything about it fail

christianbundy:
That's... very strange. Sorry that the installer isn't working correctly, that isn't good at all.

Could you walk me through what happened when you double-clicked the new installer? Did it have the usual installation workflow or did it immediately crash as "access denied"? I was under the impression that we [incorrectly] didn't actually use the AppData directory, so I'm not sure why that would break.

Does this error (and weird directory permission) persist across restarts? Thanks for your help debugging this!

ThunderFD:
I'm not 100% sure anymore but I think it crashed like it does now almost instantly:

![image](https://user-images.githubusercontent.com/17814079/56929789-e04bf580-6ada-11e9-851c-3172c53a0a80.png)

(now)

my previous version was 3.11.4 .... or so, since i wasn't logged in for a while
so it had to uninstall this first and somehow ran into issues?

I did a reboot a few hours ago and the error still persisted, yes.

christianbundy:
Hmm, these might be relevant:

- https://github.com/electron-userland/electron-builder/issues/2907
- https://github.com/electron-userland/electron-builder/issues/2992z

[...]

PR #1018, opened 2019-05-03
Fix pull-resume pagination bug

### LINKED ISSUE #978: private tab is missing messages
Maybe a paging issue? 

Might also affect other views but harder to tell. It seems that most recent messages are there, but then there is a gap and after that is older messages.

cc @ahdinosaur

christianbundy:
The problem seems to be caused by the marker counter getting ahead of
the limit constrained by `pull.take()`. This moves both the filter and
the `pull.take()` before the marker counter, which ensures that if the
counters become out of sync the marker counter can only be behind --
never ahead.

Still unclear about why this changed between scuttlebot 12 and 13. :man_shrugging: 

Resolves #978

### DIFF
--- a/lib/pull-resume.js
+++ b/lib/pull-resume.js
@@ -6,19 +6,19 @@ const extend = require('xtend')
 module.exports = {
   source: function (stream, { getResume, limit, filterMap }) {
     if (limit) {
-      let marker = { marker: true }
+      let marker = { marker: true, resume: null }
       let count = 0
+
       return pullCat([
         pull(
           stream,
+          filterMap,
+          pull.take(limit),
           pull.through(msg => {
             if (!msg.sync) {
               marker.resume = getResume(msg)
             }
-          }),
-          filterMap,
-          pull.take(limit),
-          pull.through(() => {
+
             count += 1
           })
         ),
--- a/lib/pull-resume.js
+++ b/lib/pull-resume.js
@@ -12,7 +12,6 @@ module.exports = {
       return pullCat([
         pull(
           stream,
-          filterMap,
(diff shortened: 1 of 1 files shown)


christianbundy:
Hmm... this might not be working anymore (???).

christianbundy:
Nevermind, I accidentally changed the order of the stream while debugging in my last commit -- I've amended the commit with the correct behavior, this is now good to go and seems to completely resolve the problem. :tada: 

Still concerned over why it broke in the first place...

christianbundy:
Woo! Thanks for testing, I think this is the last bug before we're finally ready for a real release!

:tada: :tada: :tada:

Powersource:
This one https://github.com/ssbc/patchwork/issues/1009 is marked as blocking tho

christianbundy:
Yep, just waiting for them to confirm whether their problem is 100% resolved on Windows. They said:

> edit: after the forced reboot I ran the installer again and everything works

But they didn't close the issue, so I'm hesitant to close it and create a release.

ThunderFD:
sorry for the delay, closed it now after trying to find info on the BSOD, but gave up^^

christianbundy:
No problem at all, thanks a bunch!

Powersource:
btw OMG I can scroll normally now :tears_of_joy: Did you update electron or something? It's buttery smooth now (as opposed to the previous molasses lol)

christianbundy:
Check out #964! It's all @nornagon, I'm super happy with the upgrade.

mmckegg:
Ya'll are doing awesome work! ❤️❤️❤️❤️❤️

PR #1020, opened 2019-05-03
Prepare for release

christianbundy:
- Re-enables `npm test` via Travis (oops)
- Updates all npm dependencies
- Removes old ssb-ebt code that wasn't doing anything

Anything else we need before a release? Should this be 3.12.0 or 3.12.0-beta?


sbot/index.js
```diff
@@ -178,20 +178,6 @@ exports.init = function (ssb, config) {
     })
   )
 
-  // update ebt with latest block info
```
mmckegg: just make sure you put this back when ebt gets enabled again 😱 😄

black-puppydog:
Running this for test, I get this:

```console
TypeError: next is not a function
    at Server.no_handler (/home/daan/code/scuttlebutt/patchwork/node_modules/ssb-ws/inject.js:20:7)
    at Server.emit (events.js:187:15)
    at parserOnIncoming (_http_server.js:652:12)
    at HTTPParser.parserOnHeadersComplete (_http_common.js:109:17)
```

Not quite sure how to trigger those though, the seem to come in batches of six, but at irregular intervals.

christianbundy:
Thanks! That must be from ssb-ws, I had it pinned to 6.0.0 (I think) because that problem was introduced in 6.1.0, but I *thought* it was resolved in 6.2.0. Guess not.

Feel free to push a commit downgrading ssb-ws or I'll do it in an hour or so. Thanks for the bug hunting!

christianbundy:
Thanks @mmckegg, that's important to remember. I made an issue for re-enabling ssb-ebt once that bug is resolved and made sure to link the code we need to re-add: https://github.com/ssbc/patchwork/issues/1021

sbot/index.js
christianbundy: Oops, replied [here](https://github.com/ssbc/patchwork/pulls#issuecomment-489331370) but I've made a note of it in [this issue](https://github.com/ssbc/patchwork/issues/1021). Thanks for pointing that out!

black-puppydog:
okay, no more of those messages, tested on two machines. seems to work for me.
On the other hand, my feed only shows very few updates. It's the weekend though, so it could just be that all the ssb folks are out and about. :P

christianbundy:
Cool, I appreciate you testing. Hopefully you're seeing more messages now? I'm seeing normal activity on my end, FWIW.

I'm thinking maybe we should release this as a beta, ask for some testing on SSB, and then plan to release 3.12.0 or something on Monday -- what do you think?

black-puppydog:
There are so many version changes, bug fixes etc in this release that having some more people test it is definitely a good idea.
I also just stumbled over something new (?) [%f5dkMKJdfRM+KEsRx8SCS1jrQkLkYuSwzIR2hR3n6dI=.sha256](https://viewer.scuttlebot.io/%25f5dkMKJdfRM%2BKEsRx8SCS1jrQkLkYuSwzIR2hR3n6dI%3D.sha256) that I didn't notice before because I'm leading the classic western "always online" lifestyle :P

christianbundy:
Sweet, once someone has time to approve this PR then I'll merge, release as 3.12.0-beta.0, and ask for folks on SSB to test it out. :+1:  

Regarding that bug: that's weird, it looks like two instances of atomic-file may be trying to write to `gossip.json` at the same time? If you can consistently repro on `master` please create a new issue (or if it's specific to `prepare-for-release` please comment on that PR!). I'll copy the same message to Scuttlebutt for posterity.

black-puppydog:
I can repro this consistently on one machine. I checked and there are no two processes. On the other machine at home it doesn't happen, so it might just be my setup.
How about we summarize what's new, and make a beta release & call for testing with a list of things that people can specifically focus on while testing?

christianbundy:
> I can repro this consistently on one machine. I checked and there are no two processes. On the other machine at home it doesn't happen, so it might just be my setup.

Interesting, I'd love if you could make a bug report so that we can refer to it. I'll merge this and plan to release a beta. The thing with atomic-file is that we use it in a few different places, so I *think* we could get a race condition even within the same process. Here's a bug I opened upstream a bit ago: https://github.com/ssbc/ssb-server/issues/597

>  How about we summarize what's new, and make a beta release & call for testing with a list of things that people can specifically focus on while testing?

:+1: Sounds good to me.

PR #1031, opened 2019-05-09
Shrink default non-expanded message size

christianbundy:
**Warning:** subjectivity! I've noticed that the default unexpanded message size is often ~1.5 screen 

## Before

Screenshot from 2019-05-09 12-47-51

Screenshot from 2019-05-09 12-48-06

## After

Screenshot from 2019-05-09 12-46-17

### DIFF
--- a/styles/base/message.mcss
+++ b/styles/base/message.mcss
@@ -108,7 +108,7 @@ Message {
   section {
     margin: 0
     padding: 0 20px
-    max-height: 1500px
+    max-height: 500px
     overflow: hidden
     -expanded {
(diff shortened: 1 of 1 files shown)


black-puppydog:
This seems *very* reasonable. But it will also mean that Dan Hassan's posts will mostly be image banners now :laughing:

christianbundy:
Ha! I'm going to try this out, maybe we can tweak in the future if it needs to be bigger? Subjectivity is hard, hopefully this is a step in the right direction.

PR #1034, opened 2019-05-11
Fix EXIF stripping via message compose input button

### LINKED ISSUE #1022: Exif data removal, feature?
Hi SSBC-Patchwork Folks,

First please excuse me if this is not the best avenue to suggest a feature, but I couldn't see any other place in the short search I did. I am in a hurry at this time, but I wanted to share this before I forget.

Anyhow,  I have noticed that photos uploaded in Patchwork still have their exif data, is the exif data removal from images something that is already in your plans?  If not, I feel it would be a nice feature to have, if possible.

Many thanks for you time reading this.

Daniel

christianbundy:
It looks like previously we were stripping EXIF data when images were
dragged and dropped into the composer or pasted, but unfortunately this
default wasn't set for the input button itself. This resolves the issue
so that when an input button is created it defaults to `{ stripExif = true }`.

Should resolve #1022, and it seems to be working correctly now that I've tested it. I'd love if someone could so we're 200% sure it's resolved. :dancing_women:

### DIFF
--- a/modules/blob/html/input.js
+++ b/modules/blob/html/input.js
@@ -11,7 +11,7 @@ module.exports = {
     return nest('blob.html.input', FileInput)
 
     function FileInput (onAdded, opts = {}) {
-      const { accept, private: isPrivate, removeExif: stripExif, resize, quality, multiple, maxSize } = opts
+      const { accept, private: isPrivate, stripExif = true, resize, quality, multiple, maxSize } = opts
 
       return h('input', {
         type: 'file',


black-puppydog:
Works for me for JPEG:

http://localhost:8989/blobs/get/&IYjh+Ze3vaEn7d8Amq8P8K4o1nVitP2H3dR/+vQkO08=.sha256?contentType=image%2Fjpeg

But **not** for PNG:

http://localhost:8989/blobs/get/&idYt6cYMMjyodKbe+mvKX+ioB9bnIHj8ZO9THq9FVqo=.sha256?contentType=image%2Fpng

christianbundy:
Sweet! Yeah, apparently only JPG/TIF/WAV have native EXIF metadata, but you can embed EXIF data in other file formats (and I think some tools do this by default if you convert from a file with EXIF).

I'd love for this sort of thing to be documented inside of Patchwork (https://github.com/ssbc/patchwork/issues/1022#issuecomment-491660814), but in the meantime maybe an entry in a changelog would be sufficient?

PR #1040, opened 2019-05-15
Remove electron-builder rules breaking AppImages

christianbundy:
During a call with @mixmix and @pietgeursen we disovered that this rule
was causing problems for AppImages on some operating systems (Ubuntu
16.04, specifically) and that removing this property solved the problem.

This increases the size of the build(s) but it's better than a broken
build. In the future we should add a new rule that decreases the build
size *without* breaking AppImages.

### DIFF
--- a/package.json
+++ b/package.json
@@ -104,9 +104,6 @@
   },
   "build": {
     "appId": "org.ssbc.patchwork",
-    "files": [
-      "!**/prebuilds/!(${os}-${arch})${/*}"
-    ],
     "linux": {
       "category": "Network"
     },


ahdinosaur:
@mixmix @pietgeursen, can you try  #1046 ?

ahdinosaur:
or if _doesn't work_, can you post up the specific error message you get?

christianbundy:
I'm going to merge this and try to see if we can get an unbroken build, would love to discuss your enhancement in your PR, I'm super excited to see whether it works!

Issue #983, opened 2019-04-06
Would it be possible to have .deb package or a repo. for Debian ?


shirishag75:
I am on Debian 10 (buster) .  I don't use .appimage but rely on .deb packages . Is there a possibility to have a .deb package or better yet a debian repo. for the same ?

Something on the lines like these - 

`deb https://riot.im/packages/debian/ stretch main`

You will need to have a public key , see this for instance 

https://riot.im/desktop.html

It would be nice if something like this can be done.

christianbundy:
Hey, I think this is a great idea. I've packaged for Arch Linux but would *love* to have this on Debian as well. Are you familiar with how to create a `.deb` package? Unfortunately I'm only familiar with making `PKGBUILD`s.

shirishag75:
I have done via using 'make' and then using 'debuild -us -uc' to make a Debian package. But for something which uses nodejs I have no clue.

christianbundy:
Yeah, my intuition is that the build will be specific to Electron, I just haven't had time to do much research  on this yet.

shirishag75:
Electron is a large package. What we have done is used riot but from a package repo. maintained by riot.im 

`deb https://riot.im/packages/debian/ stretch main`

shirishag75:
Electron is a large package. What we have done is used riot but from a package repo. maintained by riot.im 

`deb https://riot.im/packages/debian/ stretch main`

Right now their infrastructure is down, see 

https://twitter.com/matrixdotorg/status/1116304867683905537

christianbundy:
Hmm, looks like we may have an option via electron-builder: https://www.electron.build/configuration/linux.html

shirishag75:
electron-builder is cool but electron itself seems to be still somewhat immature 

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=927057

I had and have plenty of bugs using riot-web (which is based on electron) under Debian. Firefox is my go-to-browser.

christianbundy:
I'm not advocating for Electron, but it *is* the foundation of this implementation. We used to have a Patchwork-like client that ran in the browser called [Patchlite](https://github.com/ssbc/patchlite) but it hasn't seen maintenance for a long time. There's also [Patchfoo](https://github.com/ssbc/patchfoo), which doesn't use JavaScript, but I don't think that removing Electron from this project is something I have time for currently.

christianbundy:
[...]

PR #1047, opened 2019-05-20
add .deb build

ahdinosaur:
closes https://github.com/ssbc/patchwork/issues/983

- add deb as additional target for Linux builds
  - kept existing AppImage and snap defaults
- default .deb dependencies seem to be sufficient
- must have `maintainer` field, so i guess that's me =^.^=

### DIFF
--- a/package.json
+++ b/package.json
@@ -108,7 +108,17 @@
       "!**/prebuilds/!(${os}-${arch})${/*}"
     ],
     "linux": {
-      "category": "Network"
+      "category": "Network",
+      "target": [
+        "AppImage",
+        "snap",
+        "deb"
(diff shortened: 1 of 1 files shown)


christianbundy:
I love it, thanks a bunch! I'll ping you if I see any issues with the Debian build -- I'm a bit surprised that you don't need any dependencies but one way to find out. :tada:

ahdinosaur:
> I'm a bit surprised that you don't need any dependencies but one way to find out.

`electron-builder` includes dependencies by default, i don't think we need any extras because as far as i understand all the native compiled modules use their own dev headers and include their own shared libraries.

Issue #1074, opened 2019-06-07
Private message avatar pair border corners do not match up


qubist:
When private messaging someone, both users' avatars are displayed side-by-side. It looks like this:

<img width="88" alt="Screen Shot 2019-06-07 at 11 55 25 AM" src="https://user-images.githubusercontent.com/3607279/59117429-a7920e00-891b-11e9-83a1-03ac60dc07a2.png">

I think it should look like this: 

<img width="89" alt="Screen Shot 2019-06-07 at 12 01 28 PM" src="https://user-images.githubusercontent.com/3607279/59117547-02c40080-891c-11e9-93f2-5260fecb2b88.png">

Note the corners on the facing sides of the avatars.

This looks better and it also seems like someone did intend this to happen: 

[styles/dark/message.mcss](https://github.com/ssbc/patchwork/blob/c2862716ca64f372da457241399a5daf7367abe6/styles/dark/message.mcss):73
```
        a {
          :first-child {
            img {
              border-top-left-radius: 3px
              border-bottom-left-radius: 3px
            }
          }
          :last-child {
            img {
              border-top-right-radius: 3px
              border-bottom-right-radius: 3px
            }
          }
       }
```

PR #1054, opened 2019-05-22
Replace emoji symlink with real path

christianbundy:
Maybe resolves #1053? This removes the symlink, which is good, but I'm not sure how we can test this.

PR #1075, opened 2019-06-07
Match private message avatar pair edges

qubist:
Part of a fix for #1074. This fixes the issue for the dark style. there are some inconsistencies between dark and light styles, so I didn't know how to fix it there.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - Fixed emoji not rendering on Windows.
+- Fixed corners of avatar images in private messages not matching up.
 
 ## v3.12.0 - 2019-05-21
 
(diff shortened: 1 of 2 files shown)


Powersource:
Thanks for the PR :)

Hmm I think it works in conversations between 2 people, but when there's 3 or more, the people in "the middle" of the list have images with all round corners.

And if you want to, you can add an entry about this to the changelog, like [this](https://github.com/ssbc/patchwork/pull/1054/commits/6633467f910914a802bac0ad80879d81b620311c)

qubist:
I think that about does it. It would be great to have someone more familiar with the styles code look over this.

qubist:
After more testing, I have found an edge case: When I send a private message to myself I see an avatar image with rounded corners on the right side and not the left.

p

Instead, I should see an avatar image with all corners rounded.

This is caused by my avatar image being considered to have the tag `last-child`. Does anyone what code applies these tags? Searching for `first-child` or `last-child` in the codebase yields nothing

qubist:
I'm now confident this is ready for mere.

christianbundy:
This looks really great, thank you! Just merged into `master` and I can confirm that this looks great on both themes. :rocket:

Issue #1072, opened 2019-06-04
Private Messages recipient number exceeded silently ignored


cryptix:
I tried to make a private message with more then 7 people.  The preview didn't warn me and after pressing _Publish_ it stayed grayed out. I had to open the server inspector console to find this:
<img width="606" alt="Screenshot 2019-06-04 at 12 58 35" src="https://user-images.githubusercontent.com/111202/58882673-133a6800-86dd-11e9-820c-c06ae5afa2f7.png">

mmckegg:
I think this might be because it’s an off by one issue. When I implemented the warning, I assumed it was 7 recps + yourself (8 total). But it turns out it is 7 total. 

Whoops. Could also be something else. But at some point this warning did work!

mmckegg:
Yep: https://github.com/ssbc/patchwork/blob/88893c971945724444c5480c417017f6931e4bfc/modules/message/sheet/preview.js#L27

That should be `>= 7`! Easy fix for someone that wants to land their first PR!

cryptix:
Hmmm.. I guess the other thing I was concerned about: why is the error not bubbling up? Isn't that the error from the publish async? Like, would it also drop if disk was full or some other unforseen edgecase?

mmckegg:
Yeah @cryptix, that is concerning. 

It should be handled by the error handler in composer.js:

https://github.com/ssbc/patchwork/blob/88893c971945724444c5480c417017f6931e4bfc/modules/message/html/compose.js#L267

Worth investigating!

christianbundy:
Resolved by #1081.

PR #1081, opened 2019-06-10
fix isPrivate count check

cryptix:
updates #1072 by changing the preliminary check to the _correct_ hardcoded amount.


christianbundy:
Looks great, thank you for fixing this!

PR #1085, opened 2019-06-10
chore: have Travis only build `master`

christianbundy:
The previous config had Travis building pull requests *and* branches, which meant that pull requests of repo branches were done twice. Super wasteful of time and electricity. This changes the config so that we build `master` (for releases) and pull requests (for testing new code).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,6 +39,7 @@ before_install:
 
 script: npm test && npm run dist
 
+
 branches:
-  except:
-    - "/^v\\d+\\.\\d+\\.\\d+$/"
+  only:
+  - master


black-puppydog:
I was just wandering around that code and wondering why that string was like that. :)

christianbundy:
I'm still not completely sure why they added it, but it's from: https://www.electron.build/multi-platform-build.html#sample-travisyml-to-build-electron-app-for-macos-linux-and-windows

christianbundy:
My best guess is maybe people are pushing branches like `v3.13.0` instead of using Git tags? Not sure what the use-case is or why it's important that Travis doesn't build those...

PR #1086, opened 2019-06-10
chore: update node dependencies

### LINKED ISSUE #1078: Don't automatically turn ascii emoji, e.g. ":)", into image emoji
This was added some time in the last year and I'm not a fan of it. The meaning of :) vs. :smiley: is quite different to me.

For reference, most chat apps that have this feature have an option for turning it on/off. Not saying *that*'s the solution we should have, just that doing the auto conversion is often controversial.

Also worth considering is how other ssb apps render this. I'm guessing (can someone verify?) that Patchbay doesn't do this, in which case posts get quite a different feel depending on the client you read them in.

This is my opinion, what do others think?

christianbundy:
Resolves #1078 from @Powersource's awesome work in ssb-markdown.


Powersource:
Should we add a changelog entry for this maybe, even if it's in a dep?

And I vote to not call these kinds of commits 'chores', makes them sound unnecessarily boring imo :P

christianbundy:
That would be great. Feel free to add a change and / or merge, or I'll add a changelog entry in a few hours / days when I have time. My strategy so far:

- if I remember to do it, I'm just adding a changelog in the branch as I'm working
- regardless I'm going through the Git history before release and making sure we have changelog entries for all of the changes, so I'd we merge any changes without an entry in the changelog it's no biggie

I like having a changelog but I'm not sure how to integrate it into our development process. Maybe we could have a Patchwork call in the next couple weeks where a few of us can talk about this (and everything else) in more depth?

Powersource:
Yeah I can log, test, and merge tomorrow, I'm on my phone atm

And a call sounds like a good idea. I'm busy this week but after that I'm relatively flexible :)

ahdinosaur:
hm, does the latest update to `@felixriesebeg/spellchecker` require node version 10 or later? i'm now getting a compiler error when building from node 8, matches the line from https://github.com/felixrieseberg/node-spellchecker/commit/3fc0fe7eab6e8174d0710744dc2faea75301166f.

christianbundy:
Oh no! That's not great, I'd be surprised if they intentionally made a backward-incompatible change in a patch commit. We could reverse this change and / or if you can repro I'm sure they'd appreciate an issue on their repo. Not sure how to resolve that other than just reverting the upgrade.

PR #1089, opened 2019-06-11
 refactor: move files out of project root 

christianbundy:
The intention here is to make the project root a bit less busy and
stimulating so that new users can get a better feel for what's going on
in the project. This isn't perfect, but I think it's an improvement and
it reduces the number of files that the user is greeted with when
looking at the Patchwork repo.

## Changes

- move JS files from root to `lib`
- rename `modules` as `lib/depject`
- rename `sbot` as `lib/plugins`
- move documents to `docs`
- add a bunch of notes to `CONTRIBUTING.md`

## Before

Screenshot from 2019-06-11 11-35-59

## After

Screenshot from 2019-06-11 11-36-15

## Future?

- move `.vscode` configs into non-Microsofty config place
- probably merge `assets` and `build`


Powersource:
Looks good!

More suggestions (we don't have to do these, partially just checking so I understand):
* Rename `lib` to `src`
* Move the files currently (i.e. before this pr) in `lib` into something like `src/lib` or `src/util`

christianbundy:
Woo! Thanks for the review, I'm hoping this doesn't cause any merge conflicts.

PR #1090, opened 2019-06-11
fix: ensure code blocks aren't cropped

christianbundy:
This makes it much easier to read code in Patchwork. Previously you had to scroll a *ton* and only had a 300px tall window to read from.


christianbundy:
Woo! Thanks Mikey. :space_invader:

PR #1094, opened 2019-06-13
feat: improve emoji support

christianbundy:
This adds an emoji font (Noto Color Emoji, thanks @mixmix!) and revamps
the emoji rendering and suggestion systems to support a bunch of new emoji.

`:emojicode:` emojis are now displayed with a font rather than as an image, but non-emojicode unicode characters still render with the system font.

See also: https://github.com/ssbc/ssb-markdown/pull/37 (which lets us do non-emojicode unicode characters with the emoji color font!)


mixmix:
heck this is cool @christianbundy ! It's like feature tennis - I'm now inspired to implement a similar improvement in patchbay

mixmix:
p.s. _watch out_ that emoji fonts module adds 6.8MB ... I wonder how much removing those other ones removes...

christianbundy:
I'm actually kind of curious -- *should* we be bundling an emoji font? I wrote this earlier and backspaced it because [I got compound emoji working](https://github.com/ssbc/ssb-markdown/pull/37#issuecomment-501880681) (!!!), but the rest of the points still stand:

> **Decision:** do we want to include an emoji font? I'm running into a weird roadblock, here are our options:
>
> 1. Use system fonts for everything.
    - **TL;DR:** Install an emoji font if you want emojis.
    - **Good:** Simple! Small! Reasonable!
    - **Bad:** Some users might not have an emoji font, which means they won't see the emoji characters we use. This might be a problem with super old Windows machines and Linux machines where the admin forgot to install emoji fonts, but I don't know how common it is.
> 2. Install an emoji font and use it on `:shortcodes:`
    - **TL;DR:** Compound emojis treated as second-class citizens.
    - **Good:** Simple-ish! Current implementation.
    - **Bad:** Kinda sucks because now we have to bundle a font, plus we have to specify the font to use when we're using an emoji. Also this doesn't work with regular unicode emoji characters, it **only works with `:shortcodes:`**, so it means [compound emoji](https://github.com/ssbc/ssb-markdown/pull/37#issuecomment-501814965) still use the system font like in option 1.
> 3. Install an emoji font and use it everywhere.
    - **TL;DR:** Things are fine except when they're not.
    - **Good**: Simpler for users.
    - **Bad:** Because of some weird `font-family` CSS quirks, defining an emoji font means we also have to define other fonts rather than just using a generic `sans-serif`. If you set `font-family: sans-serif, NotoColorEmoji` then the computer says "funny joke! custom fonts always take precedence over generic fonts, I'm using the emoji font for everything!" and it uses the emoji font (which is like monospace with serifs? it's the worst). We can define a bunch of fonts to use *before* falling back to the emoji font, but if we run across a system without any of those downloaded (read: Linux machines where the admin installed only their favorite fonts) then it falls back to the emoji font again and everything will look like crap.

It sounds like if we bundle an emoji font it's because of people on older operating systems and Linux machines that don't have one installed yet? I'm not sure what the right thing to do is.

mixmix:
I didn't see problems with font-family: sans-serif, NotoColorEmoji in patchbay
maybe check what I did in there and see if it behaves ok on your system ?

christianbundy:
I can confirm the same behavior in all forms of Chromium (including Patchbay). Firefox works fine.

- [Chromium screenshot](https://user-images.githubusercontent.com/537700/59516351-fe489c00-8e75-11e9-9f0a-71df906f839c.png)
- [Firefox screenshot](https://user-images.githubusercontent.com/537700/59516330-fa1c7e80-8e75-11e9-8785-b3837e9c468e.png)
- [Patchbay screenshot](https://user-images.githubusercontent.com/537700/59516384-17514d00-8e76-11e9-9e57-af1e7e6f89c6.png)

I *think* what's happening is Chromium is saying "you want a generic sans-serif, I'll look for one!" and starts by checking the fonts specified in `font-family` to see whether any of them match. I think NotoColorEmoji is being identified as a sans-serif font, so it takes precedence over the *other* fallback font when it can render a character (` `, `*`, `#`, etc). This bug does *not* happen with `font-family: serif, NotoColorEmoji`.

In Patchbay you've set `font-family: arial, sans-serif, NotoColorEmoji`, but since I don't have Arial it falls back to the same bug I mentioned above:

> We can define a bunch of fonts to use *before* falling back to the emoji font, but if we run across a system without any of those downloaded (read: Linux machines where the admin installed only their favorite fonts) then it falls back to the emoji font again and everything will look like crap.

In this PR I just wrap emoji in `<span class="Emoji">🚀</span>`, and then I have some CSS setting the emoji font on `.Emoji`. That way we don't have to specify fonts that *must* be installed or have any risk that a bunch of stuff is going to break if the right fonts aren't installed. Do you think that's a good way forward, or would you recommend doing it another way?

christianbundy:
Woo! After upgrading to ssb-markdown@5.0.0 we now have consistent emoji!

Both :coffee: and `:coffee:` are rendered with an emoji font, *and* compound emoji like 👨🏾‍🎤 are rendered consistently too!

Screenshot from 2019-06-14 19-43-38

lib/depject/message/html/markdown.js
```diff
@@ -108,18 +109,3 @@ function LoadingBlobHook (hasBlob) {
     }
   }
 }
-
-function LargeEmojiHook () {
```
ahdinosaur: does this change remove the large emoji feature?

is it possible to keep this feature by changing this code to search for `p > span.Emoji:onlyChild`, add a `-large` class to matches, and in the css add a `.Emoji .-large` class with a larger `font-size` (`24px`, double the standard font size)?

ahdinosaur: as in, i like the feature of making standalone emoji on a single line large, without having to use an `h1`, since (as you've described before) that messes with the document semantics.

christianbundy: Good idea! I didn't consider how simple that would be with font-size, thanks for the suggestion! I've just pushed a commit to re-add the large emoji.

christianbundy:
Woo! :ok: :ok_man:

styles/base/base.mcss
```diff
@@ -1,11 +1,14 @@
+html {
+  font-size: 12px
+}
+
 html, body {
   margin: 0
-  font: caption
```
ahdinosaur: was this change intentional?

as far as i can tell, here's the difference:

before:

Screenshot_20190615_165153

after:

Screenshot_20190615_165118

styles/base/base.mcss
```diff
@@ -42,9 +45,13 @@ a {
 * {
   box-sizing: border-box
 }
-input, textarea, keygen, select, button {
-  font-family: '.SFNSText-Regular', sans-serif
```
ahdinosaur: was this intentional? https://github.com/ssbc/patchwork/pull/1094/files#r294034847

PR #1095, opened 2019-06-14
feat: add script to output release notes

christianbundy:
Inspired by [common-readme/cli.js](https://github.com/noffle/common-readme/blob/master/cli.js), this makes it so that our release notes are easy to generate without any manual editing. Also people can add improvements so it something I have to decide on!


christianbundy:
Context: copy and pasting this template is a pain and it's super easy to get broken links. Also I don't want to be The Person Who Decides so it seems right to decide this in the repo.

christianbundy:
I've been using this but having to switch back and forth between branches. This back-and-forth caused #1110, where I posted the wrong release notes. :scream: I'm going to merge this so that there aren't any future issues with the release notes being out-of-date with what's on `master`.  :rocket:

PR #1096, opened 2019-06-14
fix: log when Patchwork is already running

christianbundy:
Previously if you tried to start Patchwork from the command-line it
would just quit if there was already an instance of Patchwork running
somewhere. This adds some small debug output that should make it more
clear what's going on.


christianbundy:
Self-merging as trivial. :bowing_man:

PR #1084, opened 2019-06-10
feat: retry update check every 24 hours

christianbundy:
This also links directly to the latest release.

Yay, observables ftw!

PR #1098, opened 2019-06-15
fix: clear update interval correctly

ahdinosaur:
fixes https://github.com/ssbc/patchwork/pull/1084#issuecomment-502334175


christianbundy:
Thanks! I've been spending too much time with observables...

PR #1099, opened 2019-06-15
feat: render more than one large emoji

ahdinosaur:
for example, on the "Emoji Check-In Thread"s,

- %1Bsk5y+FzgNw0Zr8winKBN0HPiYWiJQB2uI+n2X+Dbw=.sha256
- %buttk1WRl2AfDp7AU1oeEvGcTnWCm7lUJ+3bVXIJqkk=.sha256

we've been posting 1-3 emojis, but our "large emoji" code only handles single emojis.

this pull request updates the "large emoji" code to handle cases with multiple emojis on the same line. which also is how other apps work, e.g. Facebook Messenger. :smiley_cat:

i tried to make the implementation as efficent as possible, but maybe this could be done better, i'm not sure.

other thread with test cases: %LAGvVzSNA/gbize/f+UeJYDEBIqWIMMDmvFZcpIY1Fk=.sha256

:heart:

### DIFF
--- a/lib/depject/message/html/markdown.js
+++ b/lib/depject/message/html/markdown.js
@@ -113,15 +113,35 @@ function LoadingBlobHook (hasBlob) {
 
 function LargeEmojiHook () {
   return function (element) {
-    // do our best with a css selector
-    element.querySelectorAll('p > .Emoji:only-child').forEach(img => {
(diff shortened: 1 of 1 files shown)


christianbundy:
Looks great! Thanks!

Screenshot from 2019-06-15 08-52-33

PR #1101, opened 2019-06-15
Use system font

christianbundy:
More notes in the commits, but the gist is that most operating systems bundle a default font that's better than what Electron uses (I think Arial?). Instead of hardcoding our favorite OS-specific fonts this lets us use the default UI font of the user's operating system, falling back to sans-serif if necessary.

Screenshot of Patchwork's homepage using a slightly different font, which is my default operating system font.

### DIFF
--- a/lib/main-window.js
+++ b/lib/main-window.js
@@ -150,7 +150,7 @@ module.exports = function (config) {
     h('style', {
       innerHTML: computed(api.settings.obs.get('patchwork.fontFamily'), family => {
         if (family) {
-          return 'body, button, input, select, textarea { font-family: ' + family + ';}'
+          return 'body, input, select { font-family: ' + family + ';}'
         }
       })
     })
--- a/styles/base/base.mcss
+++ b/styles/base/base.mcss
@@ -45,6 +45,9 @@ a {
 * {
   box-sizing: border-box
 }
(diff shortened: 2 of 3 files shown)


christianbundy:
Another benefit: when using the default `sans-serif` with different alphabets and writing systems you can get weird whitespace problems, whereas `system-ui` uses consistent whitespace around the characters. I'm not sure *why* this happens, but in my experience CJK fonts look much better with `system-ui`:

## sans-serif

Notice the lack of padding on the top of the header navigation links.

Screenshot from 2019-06-15 10-21-16

## system-ui

Screenshot from 2019-06-15 10-21-48

lib/main-window.js
```diff
@@ -150,7 +150,7 @@ module.exports = function (config) {
     h('style', {
       innerHTML: computed(api.settings.obs.get('patchwork.fontFamily'), family => {
         if (family) {
-          return 'body, button, input, select, textarea { font-family: ' + family + ';}'
+          return 'body, input, select { font-family: ' + family + ';}'
```
ahdinosaur: does removing `button` and `textarea` change anything?

christianbundy:
Woop woop, thanks for taking a look at all these PRs!

PR #1106, opened 2019-06-16
feat: re-enable ssb-ebt

christianbundy:
This was previously disabled because of a bug that killed replication,
but once that bug is resolved this should be good to merge.

See: https://github.com/ssbc/ssb-ebt/issues/22

(I don't think we should merge this until someone can confirm that the replication bug was fixed)


mmckegg:
Oh, this is just waiting on some testing? Someone should definitely do that! Merging this will massively improve initial load and sync performance.

I'll give it a try now.

christianbundy:
Doing it!

PR #1115, opened 2019-06-18
build: clarify support for Node.js Active LTS

christianbundy:
In the past the install file said we supported v6+ but Travis only ran
tests for Node 10. I think it would be great to support older versions
of Node when possible but that since we don't have many resources for
maintenance and support we should clarify that this is only guaranteed
to work on the Node.js Active LTS release (currently Node 10).


christianbundy:
Thanks Mikey!

PR #1116, opened 2019-06-18
feat: make alt text visually accessible

christianbundy:
I've seen that Dan has been adding captions to lots of his images
lately, which is super rad, but it's a bummer that he has to put the
descriptions in both the alt text *and* a secondary paragraph to make it
visually accessible.

This change takes the alt text and puts it into a `<figcaption>` that's
hidden from accessibility tools, that way screen readers see the image
alt text but people using Scuttlebutt visually are still able to see the
alt text that the community is adding to their images.

My hope is that this highlights the utility behind alt text and inspires
people to describe their visual data in a way that can be better
understood by accessibility tools.

Screenshot from 2019-06-18 09-22-42

Screenshot from 2019-06-18 09-23-26

Screenshot from 2019-06-18 09-30-25

Screenshot from 2019-06-18 09-30-44

I don't remember where I got the blue-ish caption color from, but it's somewhere else in the interface too. Maybe from "active channels" header on the left sidebar.

### DIFF
--- a/lib/depject/message/html/markdown.js
+++ b/lib/depject/message/html/markdown.js
@@ -39,7 +39,7 @@ exports.create = function (api) {
       })
     }
 
-    return h('Markdown', {
+    const output = h('Markdown', {
       classList,
       hooks: [
         LoadingBlobHook(api.blob.obs.has),
@@ -74,6 +74,22 @@ exports.create = function (api) {
         imageLink: (id) => id
       })
     })
+
+    const imgs = output.querySelectorAll('img')
+    imgs.forEach(img => {
+      if (img.alt.length === 0) {
+        return
+      }
+
+      const newNode = h('figure', [
+        img.cloneNode(),
+        h('figcaption', { 'aria-role': 'hidden' }, img.alt)
+      ])
+
+      img.parentNode.replaceChild(newNode, img)
+    })
+
+    return output
   }
 
   function renderEmoji (emoji, url) {
--- a/styles/base/base.mcss
+++ b/styles/base/base.mcss
@@ -59,12 +59,3 @@ Emoji {
   }
 }
 
-figure {
-  margin: 0
-  padding: 0
-}
-
-figcaption {
-  margin-bottom: 1rem
-  color: #527b90
-}
--- a/styles/base/base.mcss
+++ b/styles/base/base.mcss
@@ -58,3 +58,13 @@ Emoji {
     font-size: 2rem
   }
 }
+
+figure {
+  margin: 0
+  padding: 0
+}
+
+figcaption {
+  margin-bottom: 1rem
+  color: #527b90
+}
--- a/styles/base/markdown.mcss
+++ b/styles/base/markdown.mcss
@@ -63,6 +63,14 @@ Markdown {
       max-height: none
     }
   }
+  (figure) {
+    margin: 0
+    padding: 0
+    display: inline-block
+  }
+  (figcaption) {
+    margin: 1rem
+  }
 
   word-break: break-word
 }
--- a/styles/dark/markdown.mcss
+++ b/styles/dark/markdown.mcss
@@ -49,4 +49,7 @@ Markdown {
       }
     }
   }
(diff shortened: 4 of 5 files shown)


cinnamon-bun:
Great idea!!

Do the colors work well in dark mode?

Subjective aesthetic idea: indent the text 20px to visually separate it from the main body text

christianbundy:
Nope, they looked *real bad* in dark mode, thanks for the suggestion. I've used some margins to separate it out and an ultra-thin border to give some visual cohesion between the caption and the image.

Please let me know if there are any other changes you'd like to see! :rocket: 

Screenshot from 2019-06-22 10-22-22

Screenshot from 2019-06-22 10-23-21

Screenshot from 2019-06-22 10-23-39

Screenshot from 2019-06-22 10-23-56

Powersource:
Idk, as someone with normal eyesight I don't mind the captions being hidden for me, especially since I can hover on the image and see the alt-text (right?). IMO putting a description of the image below the image is annoying and mostly distracting.

Ok now I can't get that to work (maybe it's just Kas doing fancy markdown stuff sometimes). Would putting this text in hover-over text be a compromise that works for you?

christianbundy:
The text you see on hover is the [title](https://html.spec.whatwg.org/dev/dom.html#attr-title), and is meant to be separate from the alt text.

Maybe we can find a better way to do this. Here's my main drive here: 

> My hope is that this highlights the utility behind alt text and inspires
> people to describe their visual data in a way that can be better
> understood by accessibility tools.

I think it's important for alt text to be visually accessible, otherwise it can be easy to forget about accessibility and just post alt text of "IMG-2019-06-23.jpg" or whatever. That's a bummer, and it makes Scuttlebutt hard to use for tons of people. How do you think we can improve that?

Powersource:
Hmm not sure. Some ideas:

Complicated way: In the future we want (or?) a fancier way of adding images. Like where you can select compression ratio, resize and crop the image (regular apps have this most of the time, and on ssb space is even more precious). This would require e.g. a modal. This modal would include a hard-to-miss textbox that says "please enter a descriptive text to help people out".

Simpler way: Add it as another thing that appears on hover, below/above the title text box.

Powersource:
Also: I don't think people know what alt/title text really is, or how to make it in markdown (like me lol). So maybe changing the default md when uploading an image to something like ``. But then how to avoid people going with the default? Maybe remove the text if we notice it's the default? Not very clean, hmm...

cinnamon-bun:
These are my older notes from the abuse-audit spreadsheet:

> Encourage users to write alt text for images.
> * Render alt text under images for everyone, to make users aware of it
> * If there's an image upload form, put an alt text field in it called "caption" or "image description"
> * Show a nag message under the compose box if alt text is missing, or have the default alt text be "write description here"

Powersource has come up with # 2 and 3 which would help, but would be separate PRs from this one.

I quite like # 1 (this PR) because it creates a culture of remembering and supporting everyone.  Also it makes it safe to put meaning in your alt text because everyone will see it.  It could be annoying to see the alt text all the time but it feels like a good annoyance to me?

For annoyance reduction, we could
* A) show alt text in smaller font or in a fainter color
* B) add a setting to hide alt text, but it's on by default
* C) show alt text on mouse hover but do it immediately instead of using built-in tooltips, so users will see it more often.  Maybe it could float above the right side of the image to avoid changing the layout when it appears

I'm not a fan of B or C because when you're writing the alt text it would be nice to know that all your readers will read it.

Powersource:
I like C, the default system hover thing is pretty bad.

> I'm not a fan of B or C because when you're writing the alt text it would be nice to know that all your readers will read it.

Isn't the purpose just to bring attention to the text (which C would surely do)? It shouldn't be necessary for everyone to read it, right, if the text is just descriptive?

Powersource:
I'd like to clarify that I'm not just trying to be obnoxious here. I see it as a balance between supporting different forms of accessibility. This change would make stuff harder for easily distracted people like me and @ktorn

PR #846, opened 2018-08-16
Add font settings

christianbundy:
I've taken the [most common values for `font-family`](https://developer.mozilla.org/en-US/docs/Web/CSS/font-family) and added them to the settings page. This makes it easy to switch font types, intuitive to hack in custom values, and serves as a good foothold for future work on font customization.

screenshot from 2018-08-15 18-24-46

screenshot from 2018-08-15 18-24-55

screenshot from 2018-08-15 18-25-08

The only regression I've found is that the "Attach File" button has a fixed width that's cut off when I use monospace, but I'm not sure whether/how that should be resolved:

screenshot from 2018-08-15 18-25-46

PR #1120, opened 2019-06-20
Not typewriter font plz

dangerousbeans:


christianbundy:
It doesn't look like there's any description with this PR, so the context I'm bringing is that I recently switched the `<textarea>` to use the Chromium default of `font-family: monospace` instead of using a sans-serif. My perspective is that the sans-serif made writing difficult, and monospace is an improvement (see *The virtue of single spaced writing fonts* in [this article](https://ia.net/topics/in-search-of-the-perfect-writing-font) that makes writing (especially Markdown) easier.

I'd like to follow the browser defaults unless there's a compelling reason to diverge, maybe you could help me understand your position a bit more? I don't think monospace fonts are appropriate for paragraph text, but I think they're great for `<textarea>` elements.

christianbundy:
Screenshot from 2019-06-20 08-18-54

side-by-side

Screenshot from 2019-06-20 08-18-31

cinnamon-bun:
Could this be a user-controllable setting?  Or a toggle button next to the Attach File button?

christianbundy:
Absolutely, I'd be happy to add a setting for this (similar to how we have a customizable font-family). I added that in #846, I think the same approach should work well here.

dangerousbeans:
I created this because I feel that the font you are writing a new post in being different from the font of the existing posts creates a mental disconnect between those things.

Patchwork is much more the facebook/twitter replacement for a lot of people, so deviating too much from those experiences I don't think is super good. Monospace fonts feels like a more poweruser thing I would expect in Patchbay

Having this as a setting with a friendly default sounds like a great idea

christianbundy:
Oops, didn't see the edit above before I replied. We're specifying `monospace`, so it should pick your default system monospace font, not Courier. For example, mine is Nimbus Mono PS.

My perspective is that reading and writing *are* two distinct activities, especially when writing Markdown, and that there's utility in having a different experience for each. Writing with proportional fonts make sense if you're using a WYSIWYG editor, but I think they're a drag when writing code.

Anyway, this sounds like it's important to you so I'm happy to merge it. Maybe we can revisit this decision in the future if someone else brings it up. Thanks for the patch! :heart:

cinnamon-bun:
> it should pick your default monospace font

I don't think you can change that on macOS?

Maybe the font setting(s) could show all your installed fonts so you could pick the specific one(s) you want.  That would help people who want dyslexia-specific fonts too.  I'll make a separate issue for that.

masukomi:
> I don't think you can change that on macOS? - @cinnamon-bun 

Pretty sure the only way to do it systemwide is to literally edit the default font file to change the glyphs in it. So, yeah, not really a thing any normal person would or could do.

PR #1131, opened 2019-06-27
fix: don't show gatherings as private in the preview

mmckegg:
Display only bug. Looks like at some point a rogue find-replace made the gatherings preview box show all gatherings as private.

Screen Shot 2019-06-27 at 1 00 29 PM

This PR reverts the change.


christianbundy:
Woo! Looks good, thanks Matt!

PR #1137, opened 2019-06-30
Fix some zhCN translations

mwang9876:
worth mention:
- translated "pub"
- translated "patchwork" to “拼布”, which means [put together]-[cloth]
- translated "scuttling" to “快速爬行”, which means "crawling quickly"

Other than those I just fixed some word use from a native speaker's perspective.


yiwang:
Pubs to 酒馆 (sounds implies alcohol) not sure it might appealing negatively to some audience. Maybe use ”房间“ sounds more neutral?

Patchwork 拼布 is literal translation. Seems hard to get to get to something 信达雅 like "微信”, "抖音". So keep its original name as is might also be an option?

But merge anyway to advance all the addition and fixes.

mwang9876:
Thanks for your inputs!

PR #1139, opened 2019-07-02
styles: revert emoji back to old size before NotoColorEmoji

mmckegg:
This PR makes the emoji appear slightly bigger. I attempted to match the size of emojis before the Noto emoji were added.

## Before:

Screen Shot 2019-07-02 at 5 11 28 PM

Screen Shot 2019-07-02 at 5 34 55 PM

## After:

Screen Shot 2019-07-02 at 5 12 19 PM

Screen Shot 2019-07-02 at 5 34 00 PM

### DIFF
--- a/lib/depject/suggest.js
+++ b/lib/depject/suggest.js
@@ -1,7 +1,7 @@
 const nest = require('depnest')
 var addSuggest = require('suggest-box')
 var resolve = require('mutant/resolve')
-var h = require('hyperscript')
+var h = require('mutant/h')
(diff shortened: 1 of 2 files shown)


christianbundy:
Sweet! I tried to make emoji bigger in the suggestion box but couldn't figure it out, thanks a lot!

I compared back and forth and I think my personal preference is the default emoji size rather than 120%, but I can barely notice it and I don't feel strongly about it at all. Super happy to merge this, thanks a lot Matt!

PR #1151, opened 2019-07-31
Create the Patchwork Design Principles

Powersource:
Now that Matt isn't leading this project, we need a new [BDFL](https://en.wikipedia.org/wiki/Benevolent_dictator_for_life). This document is that. More details in the document.

I wrote a first draft for this document a few hours ago and then we modified it a bunch during the contributor call. But it's not done, and now we invite others to express opinions on and make suggestions for it, in line with the advice process mentioned in the document.


Powersource:
Thinking we'll make another pass with active contributors, and after that invite a bunch of others.

docs/DESIGN.md
```diff
@@ -0,0 +1,33 @@
...
+
+### Stakeholders
+* Patchwork contributors
+* Patchwork users
+* The broader Scuttlebutt community
+* People outside the Scuttlebutt community who might be affected
+
+## Design goals
+
+### Goals and scope: Patchwork should...
```
Powersource: Note to self: add something about healthy design

docs/DESIGN.md
```diff
@@ -0,0 +1,33 @@
...
+* Follow the [Scuttlebutt principles](https://www.scuttlebutt.nz/principles/)
+* Be a client that people can feel confident recommending to their friends who aren't using Scuttlebutt
+  * Polished enough usability for people without patience or skills for tech troubleshooting
+  * Safety, blocking, harassment-stopping features for people concerned about those things
+  * Strike a balance: people should know how much data they're making public; also make it hard to find some public data which could be used for harassment
```
Powersource: This line feels like it should be moved/rephrased. Maybe put it in the previous line as "privacy" or something.

docs/DESIGN.md
cinnamon-bun: How about replacing the whole section with this:

### Goals and scope: Patchwork should...
* Follow the [Scuttlebutt principles](https://www.scuttlebutt.nz/principles/)
* Be a client that people can feel confident recommending to their friends who aren't using Scuttlebutt
* Be safe (privacy, blocking, etc)
* Be usable (not confusing or broken; clear explanations of new concepts)
* Be healthy (non-addictive, calm, consentful)
* Be accessible (to the extent practical -- some a11y features are hard to retrofit into the existing code and we will build the next client with this in mind)
* Be a fun client to use -- users should enjoy using it and, in a positive way, be tempted to try other Scuttlebutt clients
* Look good

Powersource: Awesome, thanks! Added this (but removed the comment about the next client since we don't control that)

christianbundy:
(Sorry about the "License Compliance" check, I was messing around with fossa.io and it decided to use the GitHub status API without asking. Bad bot.)

docs/DESIGN.md
```diff
@@ -0,0 +1,34 @@
...
+### Goals and scope: Patchwork should...
+* Follow the [Scuttlebutt principles](https://www.scuttlebutt.nz/principles/)
+* Be a client that people can feel confident recommending to their friends who aren't using Scuttlebutt
+* Be safe (privacy, blocking, etc)
+* Be usable (not confusing or broken; clear explanations of new concepts)
+* Be healthy (non-addictive, calm, consentful)
+* Be accessible (to the extent practical -- some a11y features are hard to retrofit into the existing code)
+* Be a fun client to use -- users should enjoy using it and, in a positive way, be tempted to try other Scuttlebutt clients
+* Look good
+
```
marylychee: I guess, my question is, are we going to try make this client, the defacto default/friendly client for newbies to explore ssb for now ?? In that case could we add

* Be intuitive in user experience for new scuttlebutt users and non-tech-job centric people

perhaps related to asking shall we help in things like transitioning to messages that don't rely on markdown/ related to how might we approach the Content warnings. prioritization of these things

black-puppydog:
I like it! :)
I also like the clarification of addressing non-tech folks. The question of which one is exactly more important -- content warnings or non-markdown messages -- is already in there from what I see.

Content warnings fall under "Be safe", doesn't it?
And non-markdown features are a "please experiment in different clients" case, up until they become accepted and used, at which point they would fall under "not broken" because otherwise users would miss out on discussions which (to a non-techie) would be hard to explain since they'd just look like markdown-based discussions.

docs/DESIGN.md
Powersource: About the non-tech thing, could we maybe change that other line to something like this instead?:

> Be a client that people can feel confident recommending to their friends who aren't using Scuttlebutt, and who might not have much tech knowledge

Powersource:
Posted about this on ssb and asked for input from people `%+59cqtBvRVzfdFDbP4S/4E6xNGUWLMkt2YmngCuJ5XU=.sha256`

Powersource:
Decided on call to merge this! :tada:

PR #1152, opened 2019-08-03
Make feed deletion an opt-in setting

### LINKED ISSUE #1140: Cannot release because of shared database
Patchwork currently supports deleted messages because it uses ssb-server@15 but some other clients don't use that so I've been hesitant to release Patchwork for the past few weeks. If we release then users will encounter errors when they use another client that doesn't support deleted messages. There are a few potential solutions but they all have trade-offs and it all comes down to how we're sharing `~/.ssb/`.

christianbundy:
- update node deps
- create new setting for auto-deletion
- ensure we check setting before doing deletion
- ensure that we can't ever delete our own messages

Resolves #1140 

One weird behavior thing: the "delete blocked feeds if the setting is enabled" loop only runs when there's a changeto the blocklist, so changing this setting doesn't actually delete anyone from your database until you block/unblock *or* restart Patchwork. This wasn't my initial preference but it might actually be a good thing? I'm not sure whether or how to [elegantly] make it so that it deletes feeds immediately after the setting is enabled.


christianbundy:
Woo, thanks!

PR #1159, opened 2019-08-15
Show content warnings and allow expand/contract

cinnamon-bun:
First pass at CW support!  Only for reading, not writing yet.

This may change slightly as the spec solidifies for CWs.

When messages have a `contentWarning` field, this renders the warning as a bar above the message.  You can click the bar to expand and contract the message.  Messages start closed by default.

This reuses the existing expand/contract mechanism ("see more", "see less").

Next steps:
* Improve CSS styles and colors
* Test what happens when there's weird stuff in the CW text like newlines, Markdown, HTML, very long CWs, etc.  CWs should be treated as a single line of plain text (according to the current spec, still in flux)
* Maybe handle emoji shortcodes in the CW text
* Allow creating posts with CWs
* Add a setting to control if they're open or closed by default
* Include CW text in the fulltext index for searching purposes

Fancy stuff, later:
* Remember the open/closed state of each CW post
* Fancy settings to auto open/close CWs by author, or by substring match in the CW text, like a blocklist / allowlist

✨  This was a joint effort with @mmckegg and @marylychee  ✨

### DIFF
--- a/lib/context-menu-and-spellcheck.js
+++ b/lib/context-menu-and-spellcheck.js
@@ -88,7 +88,7 @@ function setupContextMenuAndSpellCheck (config, { navigate, get }) {
 
       if (ref.isBlob(extractedRef) && menuInfo.hasImageContents) {
         const copyEmbed = new MenuItem({
-          label: `Copy Embed Markdown`,
+          label: 'Copy Embed Markdown',
           click: () => {
             // Omit the mailto: portion of the link; we just want the address
             clipboard.writeText(`![${menuInfo.titleText}](${extractedRef})`)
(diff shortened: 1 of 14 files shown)


cinnamon-bun:
Oh right, screenshots!  This is dark mode.

Screen Shot 2019-08-14 at 5 38 14 PM



We thought that it could feel alarming to scroll past a bunch of yellow bars so we made them monochrome.  The Mastodon clients I've seen also made similar choices.  Let's try it this way first?

christianbundy:
> We thought that it could feel alarming to scroll past a bunch of yellow bars so we made them monochrome. The Mastodon clients I've seen also made similar choices. Let's try it this way first?

True! Yep, this LGTM, please merge whenever you'd like.

PR #1163, opened 2019-08-19
Add context menu option to open image in browser

christianbundy:
This finally unblocks https://github.com/ssbc/ssb-markdown/pull/29 so that our Markdown doesn't have to auto-linkify images. When a user right-clicks an image it will give them the option "Open Image with Browser", which means we'll *finally* be able to add images that link to things other than the image itself (e.g. YouTube screenshot that links to the video URL).

### DIFF
--- a/lib/context-menu-and-spellcheck.js
+++ b/lib/context-menu-and-spellcheck.js
@@ -16,7 +16,7 @@ function setupContextMenuAndSpellCheck (config, { navigate, get }) {
 
   const contextMenuBuilder = new ContextMenuBuilder(spellCheckHandler, null, true, (menu, menuInfo) => {
     const ddg = new MenuItem({
-      label: 'Search with DuckDuckGo',
+      label: 'Search With DuckDuckGo',
       click: () => {
         const url = `https://duckduckgo.com/?q=${encodeURIComponent(menuInfo.selectionText)}`
         shell.openExternal(url)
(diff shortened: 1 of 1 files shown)


lib/context-menu-and-spellcheck.js
```diff
@@ -95,6 +95,16 @@ function setupContextMenuAndSpellCheck (config, { navigate, get }) {
           }
         })
         menu.append(copyEmbed)
+        const openImageInBrowser = new MenuItem({
+          label: 'Open Image with Browser',
```
qubist: Is there a precedent in Patchwork regarding whether to capitalize the `w` in with in a context menu item? From a quick glance at context menus on various elements, I see no other uncapitalized words. Some menu items on my Mac feature a capitalized `With`:
Check Grammar With Spelling
Thoughts?

christianbundy: Thanks for the review! :eyes: If you highlight text and right-click it does "Search with DuckDuckGo", so I followed that precedent, but I'm happy to switch both to "With" if you'd prefer.

I have absolutely no preferences here, so if you're leaning one way or another please feel free to change it. :bowing_man:

qubist: Looking to Firefox for inspiration I've found `Search Google for "string..."` which, on consideration, sounds terrible because it's implying that what you're searching _is Google_, instead of that you're _using_ Google to search.

_(unless "Google" is now an adverb as well as a verb as well as the original proper noun. 🙃)_

I prefer capital-W-With, based on:
* capitalization patterns I'm seeing in the rest of Patchwork (only shorter words like `for` and `in` are uncapitalized)
* macOS 10.14.6 menu formatting (`Open With...` in finder, afformentioned `check Grammar With Spelling`)

Added commit!

christianbundy: Looks really great to me, thanks for taking care of that! It looks great. :+1:

PR #1166, opened 2019-08-20
allow "Copy Message Text" on gatherings

mmckegg:
**Problem:** right-clicking on a gathering and selecting "Copy Message Text" always throws the error _Could not get message text._

This is because the existing code assumes that all messages with message content store this on the message as `text` but for gatherings, this is done with about messages.
 
---

The PR **refactors the `getMessageText` function out of the context menu** and **adds a handler for the weird gatherings case** to look up the message title and description through _about_.


christianbundy:
This looks great, thanks for fixing this!

PR #1167, opened 2019-08-20
fix showing gatherings in feed when first created

mmckegg:
**Problem:** When you first create a gathering, it doesn't display in the gatherings feed. You have to reload patchwork before it does.

This is because the handler that sends the realtime notification to update the gatherings page is checking to see if the gathering has valid details before displaying it, but due to a race condition, the details about message has not yet been indexed, so the post gets ignored.

**The PR fixes this by bypassing this check for new gatherings created by the current user.**


christianbundy:
:rocket: :rocket: :rocket: :rocket: :rocket: 

Weird bug, I've never noticed this but I'm happy we have a fix. LGTM!

PR #1172, opened 2019-08-21
Downgrade ssb-legacy-conn to 1.x.x

### LINKED ISSUE #1170: Patchwork 3.16.0 indexes forever when you make a post
Doing anything that writes to your feed (such as liking a post) puts Patchwork into indexing mode forever.  I think the write succeeds but you can't do any more while indexing mode is happening.  Restart Patchwork and it seems ok, the write is there, but the next action will put it into indexing mode again.

Rolled back to 3.15.0 and didn't have this problem.

MacOS 10.11.6

christianbundy:
Resolves #1170 maybe?

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1458,7 +1458,6 @@
       "requires": {
         "anymatch": "^1.3.0",
         "async-each": "^1.0.0",
-        "fsevents": "^1.0.0",
         "glob-parent": "^2.0.0",
         "inherits": "^2.0.1",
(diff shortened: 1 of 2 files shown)


cinnamon-bun:
Doesn't fix the bug; turns out we need to downgrade to `1.0.12` instead of `^1.0.0`

More notes at https://github.com/ssbc/patchwork/issues/1170#issuecomment-523283152

PR #1188, opened 2019-09-20
Upgrade deps

christianbundy:
This is #1179 plus some more upgrades. Woo.

### DIFF
--- a/lib/catch-links.js
+++ b/lib/catch-links.js
@@ -16,7 +16,7 @@ module.exports = function (root, cb) {
 
     let href = anchor.getAttribute('href')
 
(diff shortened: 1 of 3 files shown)


black-puppydog:
So I've been running this branch (with conn/rooms on top) for a while now. I'm seeing some blob errors when starting without network, and git-ssb doesn't seem to work, but I'm not sure why... Apart from that it works fine.

christianbundy:
@black-puppydog Are those bugs specific to this branch? This shouldn't change anything about blob replication, as it's just an update to deep-equal, micro-css, spacetime, electron, electron-builder, and standard.

Issue #1082, opened 2019-06-10
Border radius inconsistent between themes


christianbundy:
See also: #1074 

I think `border-radius` declaration should always be set in the base theme. Here's what I'm seeing instead:

```console
$ git grep border-radius
base/avatar.mcss:  border-radius: 3px
base/base.mcss:    border-radius: 0
base/button.mcss:  border-radius: 3px
base/feed-meta-summary.mcss:    border-radius: 3px
base/feed-meta-summary.mcss:        border-radius: 3px
base/feed-meta-summary.mcss:          border-radius: 3px
base/main-window.mcss:        border-radius: 3px
base/main-window.mcss:        border-radius: 3px
base/main-window.mcss:        border-radius: 10px
base/message.mcss:    border-radius: 3px
base/message.mcss:        border-radius: 3px
base/profile-list.mcss:        border-radius: 10px
base/tag.mcss:  border-radius: 2px
base/taggers-list.mcss:        border-radius: 10px
base/toggle-button.mcss:  border-radius: 3px
dark/all-channels.mcss:    border-radius: 3px
dark/all-channels.mcss:      border-radius: 3px
dark/channel-list.mcss:    border-radius: 3px
dark/channel-list.mcss:      border-radius: 3px
dark/compose.mcss:  border-radius: 3px
dark/gathering-card.mcss:        border-radius: 3px
dark/image-input.mcss:    border-radius: 3px
dark/main-window.mcss:        border-radius: 10px
dark/main-window.mcss:        border-radius: 10px
dark/main-window.mcss:        border-radius: 3px
dark/markdown.mcss:      border-radius: 3px
dark/picker.mcss:    border-radius: 3px
dark/picker.mcss:      border-radius: 3px
dark/picker.mcss:    border-radius: 3px
dark/profile-header.mcss:      border-radius: 3px
dark/profile-list.mcss:    border-radius: 3px
dark/profile-list.mcss:        border-radius: 3px
dark/profile-preview.mcss:  border-radius: 3px
dark/tag-list.mcss:    border-radius: 3px
dark/taggers-list.mcss:    border-radius: 3px
light/all-channels.mcss:    border-radius: 5px
light/all-channels.mcss:      border-radius: 4px
light/channel-list.mcss:    border-radius: 5px
light/channel-list.mcss:      border-radius: 4px
light/compose.mcss:        border-radius: 4px
light/image-input.mcss:    border-radius: 5px
light/main-window.mcss:        border-radius: 2px
[...]

PR #1191, opened 2019-10-03
Fix border radius

foopod:
This is for #1082.

Moved everything to 5px in the base styles.


christianbundy:
Thanks for this patch, I appreciate how you've simplified the theme code to:

- manage `border-radius` in the base theme
- use the same `border-radius` everywhere

Merging now!

Issue #955, opened 2019-03-14
http_proxy support


dpwiz:
Patchwork does not appear to respect system proxy settings, thus unable to connect anything beyond local network e.g. internet pubs and dht.

christianbundy:
@wiz 

Thanks for the bug report. Could you specify which version of Patchwork, which operating system (and version), and which "system proxy settings" you mean? Thanks!

wiz:
Patchwork 3.13.0
Ubuntu 19.04
Proxy settings managed by gnome settings manager and passed through `http_proxy` environment.

christianbundy:
Is `http_proxy` a specific piece of software, or does this apply to HTTP proxies in general? Patchwork only uses HTTP for update checks and the internal blob store, so if you want your traffic proxied *in general* you'd have to proxy all TCP traffic. How does Chromium handle this? It looks like we can support a `--proxy-server` flag but that "overrides the system setting", which makes me think this should be supported natively.

wiz:
It's a way to pass the system proxy configuration to application. Then it is strange indeed that it doesn't get picked up.

As for TCP traffic there is a socks proxy, but it isn't used either and the node can't get out to a wider network.

christianbundy:
Does Chromium work correctly with your configuration, or do you have to do additional setup within Chromium?

wiz:
Chrome does pick up proxy settings correctly.

wiz:
I tried installing slack and it, too, ignores the proxy. Perhaps Patchwork could do better than that.

dpwiz:
Welp.

ajacobs-isp:
@dpwiz I believe you can set it manually when you start up Patchwork with `patchwork --proxy-server=127.0.0.1:1234` but support for "system" proxy config across Windows/macOS/Linux is a big gap in the Electron packager.

funwhilelost:
@dpwiz I believe you can set it manually when you start up Patchwork with `patchwork --proxy-server=127.0.0.1:1234` (see [Electron docs](https://github.com/electron/electron/blob/master/docs/api/command-line-switches.md#--proxy-serveraddressport)) but support for "system" proxy config across Windows/macOS/Linux is a big gap in the Electron packager.

PR #1196, opened 2019-10-24
README: adding note about proxy config

funwhilelost:
Looks like a lot of users have seen this with Electron apps so I copied the README notes from Mattermost's.  This should address #955 for @dpwiz


christianbundy:
LGTM! Do you mind if I move this to the install document instead of the readme?

funwhilelost:
@christianbundy Oh, absolutely.  Sorry I missed that.

funwhilelost:
@christianbundy I moved the block over there in this PR so you don't have to do the legwork. 🍺

PR #1204, opened 2019-11-14
Fix href bug causing errors for KIO

### LINKED ISSUE #1203: KIO error when clicking on profile
I'm running a current Arch Linux system with

* ssb-patchwork 3.16.2-1
* kio 5.64.0-1
* kio-extras 19.08.3-1

When I clicked on my profile, I got a pop-up:

```
ERROR: KIO Client
Unable to run the command specified. The file or folder /opt/ssb-patchwork/assets/@... does not exist.
```
(where "..." is a hash.)

It did then show my profile, but I have no idea what the system wanted to do in addition to showing the profile...

christianbundy:
Our `<a>` elements don't actually need a `href` because we're managing the navigation with JavaScript. This may resolve #1203, which seems to be a strange bug caused by our `href` to pages that don't actually exist.

cc: @kjcole, would love if you could test whether this resolves your issue

### DIFF
--- a/lib/depject/channel/html/preview.js
+++ b/lib/depject/channel/html/preview.js
@@ -28,9 +28,9 @@ exports.create = function (api) {
   return nest('channel.html.preview', function (id) {
     const yourId = api.keys.sync.id()
     const channel = api.channel.sync.normalize(id)
-    let href = '#' + channel
+    let target = '#' + channel
     try {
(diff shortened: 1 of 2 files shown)


christianbundy:
(I'm expecting that the Travis tests will fail until #1199 is merged.)

PR #1208, opened 2019-11-16
URL change getting-started > get-started

marcin93:
Link from message "You are not following anyone" was pointing to 404. Destination page was moved.


christianbundy:
Thanks for this patch! Looks great.

PR #1209, opened 2019-11-17
Toggle settings with CommandOrCtrl+.

jedahan:
Probably not the best way to do this, but it scratched an itch I had

### DIFF
--- a/lib/main-window.js
+++ b/lib/main-window.js
@@ -249,7 +249,7 @@ module.exports = function (config) {
     views.html
   ])
 
-  const toggleView = view => { if (views.currentView() === view) { views.goBack() } else {navigate(view)} }
+  const toggleView = view => { if (views.currentView() === view) { views.goBack() } else { navigate(view) } }
 
   const previewElement = api.app.linkPreview(container, 500)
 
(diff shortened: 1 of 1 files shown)


christianbundy:
Is this a common standard? I'm aware of `Cmd + ,` on macOS and I'd support that change, but I'm not familiar with binding `.` (or having it do both Cmd and Ctrl). Could you add some background on this change?

jedahan:
I couldn’t get that shortcut to work for some reason! I would definitely prefer , over .

jedahan:
Huh, it works now, switch to `CmdOrCtrl+,`

christianbundy:
Thanks for switching to `,`. Are you using an OS where `Ctrl+,` is common, or are you using `Cmd+,`? I'd be super down for `Cmd+,` because it's a well-known shortcut, but I think I'd like to avoid `Ctrl+,` unless that's a problem that you (or others) specifically need solved.

What do you think?

jedahan:
I'm using linux, and I think its a free-for-all but Cmd (meta) is not common but Ctrl is.

jedahan:
It might be worth looking at KDE, Gnome, XFCE, Pantheon, i3, fvwm, etc and try and follow their HIGs or our best guess at what a) does not conflict and b) seems to fit after detecting which DE/WM someone is using. For now I think this is an okay incremental improvement.

christianbundy:
Looks great to me, merging now!

PR #1214, opened 2019-11-26
Fix release notes

### LINKED ISSUE #1213: 3.17.1.AppImage does not start linux
linux Patchwork-3.17.1.AppImage  does not start

christianbundy:
This fixes the URLs generated by the release notes and adds info for
fixing the AppImage on kernels that need the `sysctl` command.

Resolves #1213

### DIFF
--- a/docs/release-notes-template.md
+++ b/docs/release-notes-template.md
@@ -2,9 +2,12 @@
 
 ## Install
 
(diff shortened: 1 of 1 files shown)


black-puppydog:
Yes, that will take care of it. As I said in #1213 and as you know, this is not a real fix, but at this point, this is definitely the way to go.

PR #1216, opened 2019-11-26
Upgrade to electron 7 and all other deps

jedahan:
I did my best to go through the changelogs or commits for all the deps.

I tested for about 5 minutes clicking around, checking markdown posts, etc. Tested only on ubuntu 18.04. The biggest three changes:

Electron 6 -> 7
ssb-markdown 5 -> 6
ssb-conn 0.10 -> 0.11


jedahan:
Tested on node 
* :heavy_check_mark:  v12 `npm install && npm start`
* :heavy_check_mark:  v10 via `npx -p node@10 npm install && npx -p node@10 npm start`
* :x: v8 via `npx -p node@8 npm install && npx -p node@8 npm start`

christianbundy:
Thanks for the patch!

Some of these were left behind intentionally in 54e41f71ebbf1f6079c89158e83cdf87abd4cc09, could you confirm that:

- **electron:** You aren't seeing "error loading sodium bindings, falling back to javascript" or similar in the CLI output
- **flatpickr:** The time and date pickers for gatherings are working again on the new versions
- **ssb-markdown:** Right-clicking an image shows an "open image" menu item (last I checked this was not working)
- **scuttle-tag:** is working 100% and you're totally sure that it's good to update (as far as I'm aware, it's abandoned code, so I've been very hesitant to touch anything)

black-puppydog:
for what it's worth, I was running this branch for a few days until I switched to andre's rooms branch just now. I didn't do active testing, but everything seemed stable.

christianbundy:
I think we need active testing on the above -- they were found as regressions when we those specific updates happened (after lots of investigation!) and so they were specifically held back in `package.json`. I'm totally open to upgrading them, but I'd like to be sure that the problems are gone.

jedahan:
* I see `error loading sodium bindings`. I wonder if we can catch and test this error / if sodium bindings are performance regressing enough to wait.
* No time & date picker shows up in create new gathering.
* I only see copy image and copy image id, no open image
* no errors with tags, but I haven't ever used them before so yeah, lets not touch them

Issue #1195, opened 2019-10-18
Pasting secret file on new computer install breaks Patchwork


jrmoreau:
I pasted my secret associated with my profile into a fresh install of Patchwork on my new computer after my last crashed. I pasted the secret into the secret file in `~/.ssb/`

Now after saving the secret file and restarting my comp the app just won't launch at all. It's on MacOS Catalina 10.5

Am I missing a step?

christianbundy:
Hey @jrmoreau, sorry to hear you're running into an issue.

Are there any other files in `~/.ssb/`, or just `secret`? I'd try starting Patchwork from the command-line, that way you'll see debug info and hopefully some useful error messages.

jrmoreau:
Thanks @christianbundy - what commands should I try? Happy to keep troubleshooting. Was resigning myself to just creating a new profile entirely.

christianbundy:
I'd personally do something like:

- Close Patchwork and don't open it until you're done
- Create a copy of your secret (and `gossip.json` / `conn.json` if you have them)
- Delete `~/.ssb`
- Re-create `~/.ssb`
- Move the file copies you made earlier into `~/.ssb/`
- Start Patchwork

jrmoreau:
@christianbundy so this is the issue - I didn't backup the gossip.json or conn.json. Only the secret because I thought it was all I needed. That's my mistake.

christianbundy:
That's fine, Patchwork should still open fine even if you don't have the `gossip.json` or `conn.json`.

Two questions:

- Are you running from the source of `master`, or one of the binary builds?
- If you're comfortable sharing it here, what's your public id (should start with `@` followed by a bunch of letters and numbers)?

jrmoreau:
@christianbundy 

- I downloaded Patchwork from https://ahdinosaur.github.io/patchwork-downloader/

- from the secret file I backed up originally my ID is @q7DgocQPPqfkwF8rvsBfdvl6bNq5XNXH4p4aGqSADyU=.ed25519

Patchwork opens fine if I open it from scratch. As mentioned, when I replace the "secret" file in ~/.ssb/ with the exact file I backed up originally, that's when it just doesn't open anymore.

christianbundy:
Are you opening Patchwork before creating the secret? You should never be *replacing* the secret file, you should be inserting the secret file into an empty `~/.ssb/` directory before opening Patchwork.

christianbundy:
[...]

PR #1218, opened 2019-12-02
Add sandbox documentation

christianbundy:
Adds info on how to start Patchwork from the CLI (#1195) and info on the sandbox problem (#1217)


christianbundy:
cc @mixmix if you want a peek at this mess

black-puppydog:
I think this makes it pretty clear. :)
The only thing that could be added IMO would be to mention that you *can* launch the AppImage with `--no-sandbox`. Of course, that requires terminal knowledge but in cases where one does not have sudo, it's good to know...

christianbundy:
Oh yeah! I'll adda note about that, I totally forgot. Thank you!

PR #1219, opened 2019-12-03
Add support for SSB Rooms

staltz:
I've been testing this branch for months for my main SSB account, it works. :) 
Now I rebased it on master.

I think we can do a better job at the UI design of room connections, but the current design isn't too bad.


black-puppydog:
This fails to build for me (with `npm ci`) with this output:

```
> electron@6.1.4 postinstall /home/daan/code/scuttlebutt/patchwork/node_modules/electron
> node install.js

internal/modules/cjs/loader.js:670
    throw err;
    ^

Error: Cannot find module 'mime-db'
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:668:15)
    at Function.Module._load (internal/modules/cjs/loader.js:591:27)
    at Module.require (internal/modules/cjs/loader.js:723:19)
    at require (internal/modules/cjs/helpers.js:14:16)
    at Object.<anonymous> (/home/daan/code/scuttlebutt/patchwork/node_modules/mime-types/index.js:15:10)
    at Module._compile (internal/modules/cjs/loader.js:816:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:827:10)
    at Module.load (internal/modules/cjs/loader.js:685:32)
    at Function.Module._load (internal/modules/cjs/loader.js:620:12)
    at Module.require (internal/modules/cjs/loader.js:723:19)
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! electron@6.1.4 postinstall: `node install.js`
npm ERR! Exit status 1
npm ERR!
npm ERR! Failed at the electron@6.1.4 postinstall script.
```

christianbundy:
This is awesome, thanks for opening the PR.

There are really only three things that jump out to me:

- I think maybe we should change the **+ Join Pub** button to be more general. Maybe "Redeem Invite" or "New Connection" or something like that?? Worse-case scenario we could do **+ Join Room / Pub
- I'd love to remove the pubs from the readme as soon as possible, because I think it makes for a super rough onboarding experience. Is there a list of nodes that implement rooms that we could reference?
- Is there a resource we could link to so that people can understand more about ssb-rooms? This is a big change and I'd really like to have documentation on how rooms are different from pubs. Is [*Announcing SSB Rooms*](https://www.manyver.se/blog/announcing-ssb-rooms) the best resource for Patchwork users?

staltz:
Thanks @black-puppydog I'll look into that.

@christianbundy 

> I think maybe we should change the + Join Pub button to be more general. Maybe "Redeem Invite" or "New Connection" or something like that?? Worse-case scenario we could do **+ Join Room / Pub

Oh good point! Maybe "+ Join Server"? I like "Redeem invite", but based on the amount of wild tutorials out there that may say "open Patchwork and click Join Pub", it might be more conservative to change the text to "Join Server" and people might still understand that it's essentially the same as "Join Pub", whereas "Redeem invite" seems significantly far from "Join Pub", that people might feel lost. Updating the tutorials is always a possibility, but who knows, maybe there's a YouTube talk as a tutorial, those are harder to update.

> Is there a list of nodes that implement rooms that we could reference?

See https://github.com/staltz/ssb-room/blob/master/FAQ.md#is-there-a-list-of-all-rooms-that-exist

> Is there a resource we could link to so that people can understand more about ssb-rooms? This is a big change and I'd really like to have documentation on how rooms are different from pubs. Is Announcing SSB Rooms the best resource for Patchwork users?

Yes that announcement on the manyverse site is a good one, but also the basic Readme for Rooms is good as well: https://github.com/staltz/ssb-room
https://github.com/staltz/ssb-room

christianbundy:
> Oh good point! Maybe "+ Join Server"?

Love it.

> See https://github.com/staltz/ssb-room/blob/master/FAQ.md#is-there-a-list-of-all-rooms-that-exist

I understand where you're coming from, but I'm very hesitant to send people on a quest to find a room server. Patchwork's readme points to a list of pubs, which I'd like to remove, but I don't think I'd be comfortable removing the quick and easy pub list with a list of search engines to try.

Note: I might be biased because I tried that list and after trying every option I found one server.

- **Twitter:** you posted about `http://80.240.20.193/`, but that isn't a valid invite so it doesn't work. Typing it into a browser also doesn't work because the server is down.
- **Facebook:** "This page isn't available. The link you followed may be broken, or the page may have been removed."
- **Instagram:** "0 posts"
- **Mastodon:** "The page you are looking for isn't here. "
- **LinkedIn:** "Make the most of your professional life `<login form>`"
- **Minds:** "Login to Minds `<login form>`"
- **Tumblr:** "We found nothing. Here it isn’t."
- **Pinterest:** Your link points to the word *candy*, but changing it to "ssbroom" just shows you [living rooms](https://www.pinterest.com/search/pins/?q=%23ssbroom).
- **GitHub:** It's just https://github.com/staltz/ssb-room
- **Google:** http://159.65.63.184/
- **DuckDuckGo:** http://159.65.63.184/
- **Bing:** http://159.65.63.184/

I'm thinking of room servers like DHT announce servers, so I'm much more comfortable with a list of known ssb-rooms nodes, but I'm open to new information. Are there concerns you have about the centralization of rooms that doesn't apply to DHT announce servers? Or would you feel the same about DHT announce node lists? I want to make sure I really understand your position fully.

black-puppydog:
thanks @christianbundy that fixed it! :)

staltz:
> send people on a quest to find a room server

I think I wrote a lot about this topic on SSB but I can't find the relevant cypherlinks right now. I'm very against the idea of a centralized list in a wiki or something, because it's centralization and brings all the problems of centralization with it, most concerning problem is that of power and moderation. For instance, [this issue](https://github.com/ssbc/ssb-server/issues/690) even though both of us agree with the outcome, is still power dynamics at play. The convenience of a centralized list is attractive, but it has a downside.

> Are there concerns you have about the centralization of rooms that doesn't apply to DHT announce servers? 

You mean bootstrapping servers? There is an important difference between rooms and DHTs: bootstrap servers introduce clients to a world-wide cyberspace with potentially thousands/millions of peers, but a room server introduces you to... one room.

So a room is convenient to encapsulate a small amount of people (probably less than 100). This makes a room ideal for small communities, and because it has a similar size to the Dunbar number (150), it serves really well for one community at a time. Now if you make all rooms public to the wide internet on a centralized list, people will join the first room that has space, losing the semantics of one community per room. That's what they do currently with pubs. Only if all (or most) rooms have space will someone joining SSB pick a room of their interest first. So rooms only scale if they are utilized well within their limits, which means that rooms will not scale well if we cross that limit (which will easily happen if there are more than 100 people interested in the same topic).

Now let me talk about the one type of centralization that is awesome: hashtags! A hashtag centralizes an interest into one word, and people can *come together to the same word* in order to find each other. But at the same time, a hashtag is decentralized concerning power, moderation, ownership. No one can own a hashtag, no one can force others (blocklist) to stop using a hashtag. This is why I'm betting on hashtags. If you want a centralized list of (public) rooms, then the hashtag `#ssbroom` centralizes that, and you can use this token in various platforms and spaces in order to find the actual room servers. And it's not *that* hard to use one of the Web search engines to find (public) rooms, it's literally one click away. I'm sure there are other rooms, but those are private (by URL or IP). I think we could remove some of these services, like Pinterest, but the idea remains: use one hashtag instead of a moderated list.

Nothing stops people from building such list, though, and I'm sure that they will if this idea grows more. But it's worth starting it the right way, without any centralized moderation. Also, Patchwork could moderate such a list, but other apps might not. E.g. in Manyverse I don't plan to have any list.

staltz:
Updated to have "**+ Join Server**" button.

christianbundy:
Oops, I think you removed my commits. I'll push another to get the tests passing again.

Here's my concern: if we can't make rooms easy to use then people will continue using pubs, which centralize data, ruin the social graph, and aggravate the rando problem by putting everyone within replication distance of each other. I think we all agree that follow-back pub invites are sub-optimal, and we're all interested in deprecating them.

I'd like to avoid making the onboarding harder, but I'd also like to try to make sure that you and I are aligned. I'd feel bad about using your awesome work and then making a list that you specifically asked me not to.

I understand that maintaining a list of Good Rooms is wielding power, and that we *do* use that power to make sure everyone we link to is friendly and aligned, but I'm not sure that's worse than saying "look up this hashtag". My major concerns:

- **We still have to choose a list of Good Search Engines**, which is a higher-order way of wielding power. Instead of saying "here are some rooms we like" we're outsourcing that trust to Google, Facebook, Twitter, etc., some of which are super vulnerable to bot activity. I'd imagine that if the Twitter results were 99% Nazi bots advocating for their Nazi rooms that we'd second-guess linking there. If someone asks you to add Voat, Stormfront, etc., you'll *still* have to choose which websites you like (and more importantly: how they're ordered).
- **Most people joining are already interested in SSB**, which means that they already have a common interest when they join the network. I feel very strongly that we need better tools for filtering content to avoid the SSB developer talk, but I don't think that the way to do that is to put people in rooms about *anything other than SSB*. Joining a room doesn't put you in follow distance, so I don't have a problem with big rooms that only aid replication + feed discovery. Click people who are online, follow them if they're cool, and you've joined the network without ruining your follow graph.
- **Onboarding is already hard enough**, and I think that the negatives we're seeing from the follow-back pubs are happening because they clobber the social graph, not because they're in a centralized list. I don't imagine that there will be a big difference between "use a few of these very popular rooms" and "use a ton of these moderately popular rooms", since the end result is a bunch of tunneled connections anyway. 

My take-away is that unless rooms are super easy and accessible, we'll have to keep recommending pubs, which I think is 100 times worse than the worst-case scenario from rooms. I share your vision of what the perfect system might look like, but I don't want "perfect" to get in the way of "better".

> Update: actually, a hashtag-based DHT room would be really awesome

Yeah! I think we chatted about this a while ago when I was investigating discovery-swarm and trying to see if it was something we could staple onto the clients.

staltz:
> Oops, I think you removed my commits. 

Ouch! I didn't realize. Sorry.

> Here's my concern: if we can't make rooms easy to use then people will continue using pubs... aggravate the rando problem

Fair point :thinking: 

> I'd feel bad about using your awesome work and then making a list that you specifically asked me not to.

So here's a misunderstanding, let me explain. Don't worry about the length of this following text, I'm just trying to communicate richly. I don't believe in lists, **but** *because it's impossible* for me to prevent the internet at large from making such lists, I won't condemn anyone else from making lists. It's actually untrue that I specifically asked you to not make lists. I didn't! :) In fact, if there would be any list of rooms, I'd rather have you and other butts do it than (say) some stranger who advocates for `<insert here some opinion which I find toxic/>`. This is part of our "supportive disagreement" mantra. 

> *We value disagreement when it’s supportive, and see it as generative and bond forming.* — https://scuttlebutt.nz/docs/principles/

All I'm saying is that I myself won't maintain and promote lists. It's like not giving it an upvote, while simultaneously not giving it a downvote either. Perhaps it's hard to grok what my motivation is, so I'll put it like this: I have opinions (e.g. against centralized list) but I don't force others to have the same opinions, and I expect them to not force their opinions onto me either (e.g. in favor of lists), **because** I know that I have no power (and shouldn't!) in dictating what people should do, because given the current conditions of the internet and the scuttleverse, anyone who truly wants to build their opinion, can do it. I have embraced the fact that being dogmatic is totally fruitless, and prefer instead to influence, and whenever influence is insufficient, I support the disagreement and want to see what awesome things they will build with their own opinions and vision.

So if your take-away is to make rooms super easy and accessible, please be my guest and go ahead with that! Note, though, that it doesn't mean I'll accept PRs to ssb-room, but I'm happy if you make a fork, I might even tweet it or promote it. This is basically how a decentralized system should grow, not with a centralized product designer, but through unconstrained swarm experimentation.

-----

On hashtag-based DHT rooms: yep they are quite doable and quite exciting. I want to do it, but between all the other tasks to do in manyverse, I'm sad I can't find the time.

christianbundy:
I'd like to write a longer post, but I want to merge this before I go to bed so this might be short.

Thanks for explaining your position, I feel like I understand it much more now. I'll reach out to some Patchwork butts regarding the pub / room list recommendation in the readme and release notes. If you'd like to be a part of that discussion I'd love to have your feedback, but I know Manyverse and SSB maintenance is keeping you very busy.

Thanks a bunch for all your work on ssb-rooms, I'm excited to have this merged.

PR #1220, opened 2019-12-03
intl: update spanish

gmarcos87:
Updates the Spanish language to 100% and removes obsolete translations.


christianbundy:
Beautiful! Thanks so much for this patch.

PR #1228, opened 2019-12-27
Change to throttle sidebar UI on public page

### LINKED ISSUE #1215: "Possible connections" glitches
Two things:

- My local pub, libra.fraction.io, keeps popping up at the top of the "possible connections" list every second and then *immediately* disappears. This makes it impossible to click and very visually distracting because it's making the others in the list move up and down in a stuttery fashion once per second. I think "pub.evbogue.com" and "Ql06HdDq1" have the same problem, but they take a few moments to show up again.
- When I hover "ansuz" and click it registers the click and the pub leaves my screen for a moment, but then it comes back. I'm not seeing any error output anywhere, but I'm guessing a connection is failing somewhere?

cc: @staltz

christianbundy:
When connections change quickly the sidebar can flash and strobe, which
is annoying at best and inaccessible at worst. This commit throttles the
UI changes at a maximum of 1 change per second, which reduces the
unwanted behavior.

Note: The `throttle()` method from Mutant doesn't have tests or
documentation, so this is my best guess at how it's meant to be used
from reading the source code.

Resolves #1215 

See also: `%wkbFpz2U4Gv/8eJxszE7ke6cnb8GMVkYjLceOn0kbe8=.sha256`

### DIFF
--- a/lib/depject/page/html/render/public.js
+++ b/lib/depject/page/html/render/public.js
@@ -1,5 +1,7 @@
 const nest = require('depnest')
-const { h, send, when, computed, map, onceTrue } = require('mutant')
+const { h, send, when, computed, map, onceTrue, throttle } = require('mutant')
+
+const slow = (input) => throttle(input, 1000)
 
 exports.needs = nest({
   sbot: {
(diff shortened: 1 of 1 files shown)


staltz:
The code looks good. Does this work as intended? If so, good to merge in my opinion.

christianbundy:
Yeah, although Bob mentioned in the linked thread that there's still an issue where Lynn movies from "Local" to "Connected" repeatedly, which probably isn't what we want?

PR #1229, opened 2019-12-30
Add ia32 build target

christianbundy:
Let's see if this builds?

cc: @soapdog


soapdog:
@christianbundy it looks like it works! Thanks a lot! :D

PR #1235, opened 2020-01-08
New architectures: ia32 and arm64

soapdog:
New architectures: ia32 and arm64 for both Linux and Windows.

Also added a new build target for windows: "zip" which makes it easier to work with package managers.

### DIFF
--- a/package.json
+++ b/package.json
@@ -142,13 +142,6 @@
           ]
         },
         {
-          "target": "nsis",
-          "arch": [
-            "x64",
-            "ia32"
-          ]
-        },
-        {
           "target": "zip",
           "arch": [
             "x64",
--- a/package.json
+++ b/package.json
@@ -115,15 +115,15 @@
       "target": [
         {
 					"target": "AppImage",
-					"arch": ["x64", "ia32", "arm64"]
+					"arch": ["x64", "arm64"]
         },
         {
 					"target": "snap",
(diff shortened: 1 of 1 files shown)


christianbundy:
Thanks for this patch!

Weirdly enough, the Linux build failed -- I think while building the spellcheck library? I've restarted the build, hopefully it will pass this time. :sparkles:

christianbundy:
I have a bad feeling about arm64 support in our dependencies.

soapdog:
I've changed the number of platforms being built. Now it only builds `arm64` for AppImages which I know kinda words because @clehner was able to run an Applmage generated this way on his pinebook.

The spellchecker might be causing trouble. I don't know I will double check the build log after it attempts building it.

soapdog:
@christianbundy this is really odd, the spellchecker failed building for x64 on Linux which is the most standard thing ever and you had it working before...

christianbundy:
I have a hunch that the x64 error is a red herring and that it's actually failing while trying to do the 32-bit Linux support. Electron has [deprecated 32-bit Linux support](https://electronjs.org/blog/linux-32bit-support) as of Electron 4, so I think it might be okay (although unfortunate) to remove that option.

soapdog:
yay, it passed the checks!

christianbundy:
Thanks for this patch!

PR #1241, opened 2020-01-20
update ssb-conn to 0.12.0

staltz:
Updating ssb-conn from 0.11.3 to 0.12 [tweaks the scheduler](https://github.com/staltz/ssb-conn/compare/014c407cdaa722e9a04906a1678b463dc06356b6..v0.12.0#diff-0b5f8eafe190c56c8bc6564584ca24d6) with a few improvements:

- Connect to slightly less staged friends
- Connect to less rooms (from 10 to 5)
- Treat rooms as non-syncable peers

Note though that Patchwork was using ssb-conn 0.10.3 so it will also get these long-due updates:

- Dont use deprecated ssbRef.parseAddress in scheduler 
- `has-network` => `has-network2` ([removes some weird unnecessary module.exports hacks](https://github.com/staltz/has-network2/commit/0da8a9157092ea09d12122a6dc1ea18debbb2659))
- `on-change-network` => `on-change-network-strict` ([uses JavaScript "use strict"](https://github.com/staltz/on-change-network-strict/commit/261163acdf330d828db3f57429e5d23e2a6102ae))
- Scheduler: update grace period for connecting peers
- Remove hacks for DHT addresses in the (legacy compatibility) gossip plugin 

Other transitive dependencies updated by this PR:

- `promisify-tuple`: 1.0.0 => 1.0.1
- `pull-ping`: 2.0.2 => 2.0.3
- `ssb-conn-hub`: 0.2.4 => 0.2.7
  - make sure all the rpc.close calls have callbacks 
  - improve inferredType for DHT peers 
  - remove hacks related to DHT addresses
- `ssb-conn-query`: 0.4.1 => 0.4.4
- `ssb-room`: 1.0.0 => 1.1.3
  - client plugin: gracefully close RPC connection with server 

**This PR was tested in production locally on my laptop, I ran Patchwork with these updates and it worked fine.**

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3550,10 +3550,10 @@
       "integrity": "sha1-tdRU3CGZriJWmfNGfloH87lVuv0=",
       "dev": true
     },
-    "has-network": {
+    "has-network2": {
       "version": "0.0.1",
-      "resolved": "https://registry.npmjs.org/has-network/-/has-network-0.0.1.tgz",
-      "integrity": "sha1-Pup7RMqpYBeXEkvouonSKMQQFJk="
+      "resolved": "https://registry.npmjs.org/has-network2/-/has-network2-0.0.1.tgz",
+      "integrity": "sha512-cpqaHIBwSKOhHzpWPXS2FnOzN6ekF238MxonA1CWZNx7UivVlq3GF+GqS6gXbvDwMYoaxXM+8el7tP8uFSrjsQ=="
     },
     "has-symbols": {
       "version": "1.0.0",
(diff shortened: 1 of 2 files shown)


christianbundy:
Thanks @staltz!

Very happy about the `'use strict'`. In case you didn't see, I tried to take care of that for all of our dependencies and tracked them here: https://github.com/ssbc/ssb-server/issues/683

staltz:
> there's a slightly weird effect when manually disconnecting from everything at the same time. But I'm guessing that was there before.

It was probably there before. I think unfortunately to fix this properly requires a rework of the connections tab. That's if I interpret "weird effect" correctly as what's in my mind.

@christianbundy This fixed just one or two cases of `'use strict'`. Would be good to fix all others. You may notice that it's a fork package, so it bothers Dominic *et al* less, but it's also a problem that there's a fork. Might try to get it merged into upstream at some point.

christianbundy:
> Would be good to fix all others. You may notice that it's a fork package, so it bothers Dominic *et al* less, but it's also a problem that there's a fork. Might try to get it merged upstream at some point.

I think forking is the best solution, see [this comment](https://github.com/ssbc/ssb-server/issues/683#issuecomment-515750552). I really wish forks weren't such a hassle, I keep thinking about how cool it would be if we had lots of closely-related forks with slightly different answers to problems like:

- which patches are merged
- npm or yarn
- javascript or typescript
- do we care about strict parsing
- do we care about style guides
- do we care about linter rules
- do we want experiments or stability

I think npm limits us here a lot, but I keep thinking that there might be a simple way to do it with Git maybe a [Git-based package manger](https://github.com/christianbundy/git-npm). :thinking:

Issue #1240, opened 2020-01-18
Settings keyboard shortcut is now global on macOS


miikka:
Whenever I have Patchwork running in the background and I press `Cmd+W` in some other application, instead of that application doing whatever `Cmd+W` does, Patchwork goes to the settings view. What I expected is that the active application handles `Cmd+W` and Patchwork doesn't do anything.

* The intended Patchwork keybinding may be `Cmd+,`. I'm using Dvorak keyboard layout, but some applications handle keybindings as if I was using the Finnish layout which is what is physically printed on the keys. 
* I'm running Patchwork 3.17.2 on macOS Catalina 10.15.2

christianbundy:
Maybe a local shortcut would solve this? https://electronjs.org/docs/tutorial/keyboard-shortcuts

miikka:
Seems like a good idea. The `Cmd-,`/`Cmd+w` mixup is a probably bug in Electron: https://github.com/electron/electron/issues/19747

Gozala:
I noticed that patchwork seems to steal all the keyboard shortcuts whenever it's running so most keyboard shortcuts on other apps like Firefox stop working. I'm pretty sure this is recent regressions as it did not used to be the case before

christianbundy:
I'd be comfortable merging a patch that used a local shortcut *or* reverts the patch that added this behavior.

PR #1243, opened 2020-01-27
Make the settings keyboard shortcut local instead of global

miikka:
See https://github.com/ssbc/patchwork/issues/1240 for background.

To make a keyboard shortcut local in Electron, it has to be put in some menu. I put it under _Navigation_ below a separator, with the idea that maybe some other views could be added there as well.

Näyttökuva 2020-1-27 kello 17 47 04


christianbundy:
Thanks so much for this patch, looks great!

PR #1246, opened 2020-02-12
add DHT invites, only as a client

staltz:
Implemented! **But** I think we need some thorough testing before merging. I myself was not able to connect Patchwork to Manyverse via DHT invites. Sometimes it depends on the NAT gods of the internet.

And note that because Patchwork doesn't display "connecting" states, only "connected" states, as soon as you press "accept invite", the UI will give you no indication that the invite was converted into a "connecting" status, so the user might feel confused that *nothing is going on*. Hopefully someone else could chime in and do something about that. Feel free to update this branch with new commits


lib/server-process.js
```diff
@@ -34,8 +19,7 @@ const createSbot = require('secret-stack')()
   .use(require('ssb-about'))
   .use(require('ssb-private'))
   .use(require('ssb-room/tunnel/client'))
-  // .use(require('ssb-dht-invite')) // this one must come before dhtTransport
-  // .use(dhtTransport)
+  .use(require('ssb-dht-invite'))
```
ciex: I always wondered why the chaining order is like it is here as it is not arbitrary, right? It would be really great to add some documentation about that here at some point.

christianbundy: It's not arbitrary, but it's also not documented. To be candid, I think the SecretStack plugin system is an anti-pattern that encourages global mutable state. I'm doing my best to randomize the plugin order in Oasis with this pattern (https://github.com/fraction/flotilla/blob/master/index.js) but it's still a mess. I really wish these modules would just `require()` their dependencies instead of using another dependency system.

christianbundy:
@staltz Is this fine to merge? I'm not worried about whether it works super well, even 10% reliability is better than 0%.

staltz:
@christianbundy Yes. Widespread testing would help us know, I only did a local test here with my few devices (all under the same network, or in the same city cellular network).

This PR needs a git conflict resolution before merging though

lib/server-process.js
staltz: > To be candid, I think the SecretStack plugin system is an anti-pattern that encourages global mutable state.

:scream: hot take! I actually like the secret-stack plugin system, out of all the other ideas in the SSB stack.

staltz: If y'all want some insight on what depends on what, I wrote it down for Manyverse here: https://github.com/staltz/manyverse/blob/3a4e80d151fe70cbd7d9fcfe133b1eec863aef14/src/backend/ssb.ts#L57-L94

Issue #1256, opened 2020-03-03
UI: Avatar images are stretched


johanbove:
As mentioned [here](https://viewer.scuttlebot.io/%25m1%2FVR8PhIRWIOLlX%2FLStyxtNX%2FgE2gz9Ed1ltWXL7o0%3D.sha256)

black-puppydog:
Game, set, match! Props to @johanbove :)

johanbove:
I seem to have missed a couple of places where the "avatar" image still looks stretched. Will look into this later.

PR #1257, opened 2020-03-03
Fixes issue #1256

johanbove:
Setting the object-fit to `cover` for square images avoids stretched images.


christianbundy:
This looks great, thanks for the patch!

Issue #1254, opened 2020-02-26
Bug: Too many recipients on 7 recipients, error message text is confusing our many selves


hexagon6:
I get the following message:

>**Too many recipients**
Private messages can only be addressed to up to 7 people. Your message has 7 recipients
>Please go back and remove some of the recipients.

![ssb_too_many_recipients](https://user-images.githubusercontent.com/6619032/75328287-10406280-587e-11ea-877c-e14cc31fd626.png)

When writing a private message and adding exactly 7 people, I cannot conclude the process and send the message. (I first had 8 recipients, got a correct error message text, removed one recipient and the got the error message above).

Maybe removing a recipient is not registered or the "form" is not properly tested on updates, or simply an off-by-one error?

hexagon6:
This happened on: ubuntu GNU/Linux with GNOME
with Release: Patchwork-3.17.5.AppImage

Powersource:
It's because you yourself count as one of the recipients. Yeah it's kinda dumb :P But it's so you can read the messages yourself.

But yeah this error text should definitely be changed.

black-puppydog:
my understanding is that this has been fixed in #1259. closing for now, but feel free to reopen if I got this wrong

PR #1259, opened 2020-03-06
Fixes issue #1254

hexagon6:
Adjust too many recipients message text to include ourselves


christianbundy:
Thanks for this patch!

PR #1268, opened 2020-03-21
Added nl locale

### LINKED ISSUE #1264: i18n: Add Dutch (NL) to locales
We're a small language in the World. About 24million people are speaking Dutch. Nevertheless it is my mother-tongue, so it deserves a place in the "language choice" menu. Should be an easy translation as it is similar to German and English.

johanbove:
Resolves #1264


christianbundy:
Looks great, thanks for this patch!

Issue #1273, opened 2020-04-04
French locale missing some translations


vtexier:
Using Patchwork 3.17.6

In the french GUI: 
* If I "Like" a message ("Aimer") the message become "**Liked**" not translated to "**A aimer**".
* Writing a new message display "**new**" at the top right corner, not translated to "**nouveau**".

black-puppydog:
I'm not really qualified to make these changes. 
But the format for the file is *really* simple. It's just the file `locales/fr.json` and with #1275 most of the yet to be translated strings are in that file, and just need to be changed.

If you want to you can just edit them directly in the github web UI. In that case I'd just need to verify on my machine that the translations still work before we can merge the changes. 
Or I quickly coach you into how to test this locally if you prefer. :)

black-puppydog:
(je peux aussi essayer de faire le coaching en français. :) )

vtexier:
Catched missing translations again:
* The friend options are not translated.
![patchwork_menu](https://user-images.githubusercontent.com/2174016/80523704-c6afe900-898e-11ea-8017-d93aca16f6d3.png)

* The age of a message is not translated:
![patchwork_age](https://user-images.githubusercontent.com/2174016/80523740-d92a2280-898e-11ea-9687-792ddfe3c5fd.png)

black-puppydog:
Something of a standing issue. Feel free to send patches. :)

PR #1275, opened 2020-04-22
update French locale.

black-puppydog:
Got triggered by #1273.
But there's a bunch of them still missing. at least now they're
explicitly in there for anyone to change.


black-puppydog:
superseded by @vtexier's PR

Issue #780, opened 2018-04-21
Mouseover of profiles should include some of the user bio


celesteh:
The who to follow side bar is interesting, but mousing over a particular suggestion only gives me their profile picture and name. Including a bit of their bio/description would make this feature more useful and I might start using it to figure out who to follow.

Cornelius-Sheahan:
This would be my first contribution to this project so I'd like to start with something small like this.

I assume the best place would be between the Name and Public Key, or is there a alternate spot?

How much of the Description should be shown?
- split on the first new line ('/n')
- character limit approx equal to the public key 
- the smallest between these first two 
- the entire length

christianbundy:
For reference, here's how it currently looks:

![Screenshot from 2019-05-05 14-46-42](https://user-images.githubusercontent.com/537700/57200776-9e8cd600-6f44-11e9-9a07-a6ced2b73852.png)

I think the description would be cool, and I think this is the right heuristic:

> - split on the first new line ('/n')
> - character limit approx equal to the public key
> - the smallest between these first two

I'd also be interested to see some basic friend info, like "blocked by 4 people you're following and followed by 42".

mmckegg:
It currently looks different if you don’t follow the person. 

You’ll see number of mutual friends or if people block them.

lluisemper:
I am willing to help, what should I do? a fork of it, implement it and merge it?

christianbundy:
Yep! There's more info in the [contributing guide](https://github.com/ssbc/patchwork/blob/master/docs/CONTRIBUTING.md).

Powersource:
@lluisemper If you want help getting started with the implementation we could set up a call and I could sort of show you the ropes :)

lluisemper:
Very kind your offer, let me a couple of days to try it out by myself and If I can not figure that out I let you know :) @Powersource 

Kind regards

Lluis

black-puppydog:
Not sure if you all got notifications for this but #1277 does something like this and seems to work fine for me locally, but I wouldn't want to merge it without at least someone else checking it out to see if it runs. :P

black-puppydog:
@celesteh I hope this version does something along the lines of what you thought when you started it. I'll be happy to iterate on this. I've been using this myself for a few days now and have to say even though it's a bit hap-hazard, I don't want to miss it any longer. :)

PR #1277, opened 2020-04-23
Add super-short bio summary to mouseover popup.

black-puppydog:
This addresses #780 in a relatively hap-hazard way.
I basically just tried to copy what the code around this feature was doing. 
Also I didn't know any other way to ditch markdown syntax than to include a new dependency.

Alternatively we could just pass the raw markdown, and hope that users will notice that it doesn't look so great, and just have super-short bios in their profiles' first lines. :stuck_out_tongue:

### DIFF
--- a/lib/depject/profile/html/preview.js
+++ b/lib/depject/profile/html/preview.js
@@ -4,9 +4,11 @@ const map = require('mutant/map')
 const when = require('mutant/when')
 const computed = require('mutant/computed')
 const send = require('mutant/send')
+const removeMd = require('remove-markdown')
 
 exports.needs = nest({
   'about.obs.name': 'first',
+  'about.obs.description': 'first',
   'about.html.image': 'first',
   'keys.sync.id': 'first',
   'sheet.display': 'first',
@@ -25,8 +27,10 @@ exports.create = function (api) {
   const plural = api.intl.sync.i18n_n
 
   return nest('profile.html.preview', function (id) {
+    const idLen = id.length
     const name = api.about.obs.name(id)
     const contact = api.profile.obs.contact(id)
+    const description = api.about.obs.description(id)
 
     return h('ProfilePreview', [
       h('header', [
@@ -42,7 +46,15 @@ exports.create = function (api) {
           ]),
           h('section -publicKey', [
             h('pre', { title: i18n('Public key for this profile') }, id)
-          ])
+          ]),
+          h(
+            'section -profile', [
+              computed(description, (description) => {
+                const txt = removeMd(description)
+                const summary = shortenDescription(txt, idLen)
+                return summary
+              })]
+          )
         ])
       ]),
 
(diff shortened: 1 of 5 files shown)


black-puppydog:
@christianbundy have you had trouble like this with travis before?

christianbundy:
Yeah, usually after a few hours you can restart the build and they'll go green again. :roll_eyes: 

I tried one restart a couple of hours ago, I'll try again later today.

black-puppydog:
Okay, best I can tell, there's an issue with how we handle caching in travis:

* It works when I disable caching entirely.
* It fails again if I re-enable it and it uses the caches it was using before because the uncached build didn''t modify those.
* It works if I put `npm cache verify` into the `before_install` hook, which produces some interesting retults:
    * On Linux (which was failing before) it shows this: `Corrupted content removed: 1`
    * On Windows, caching seems to fail: `Content verified: 0 (0 bytes)`. After the build, *uploading* the cache seems to fail, which would explain that. `failed to upload cache` is not a reason for travis to fail the build though.

If nobody else wants to have a go at this, I'd say we leave it like this for now, at least if it breaks now we might get more pointers as to what's wrong.

black-puppydog:
So, travis problems notwithstanding, I'm still unsure about the quality of this PR and would like input. :P

christianbundy:
I'd be happy to help, but I think I need to take a few days away from Patchwork and try to avoid burnout. :/ I'm happy to look at this next week though, or if anyone else has energy then please feel free to merge it.

black-puppydog:
Yes, of course Christian! This is by no means urgent, so you just make sure you're good before touching this with a ten foot pole. More important now than ever. :)

lib/depject/profile/html/preview.js
```diff
@@ -4,9 +4,11 @@ const map = require('mutant/map')
 const when = require('mutant/when')
 const computed = require('mutant/computed')
 const send = require('mutant/send')
+const removeMd = require('remove-markdown')
```
Powersource, line 7: Instead of this, and `shortenDescription`, I think you should be able to use https://github.com/ssbc/patchwork/blob/master/lib/depject/message/async/name.js#L53 instead. It's what we use to create the text for e.g. links to forks and references.

black-puppydog:
Okay @Powersource, I think I have sth usable here. The markdown summary function was a helper in a depject module. But I read [somewhere](https://github.com/ssbc/patchwork/blob/master/docs/CONTRIBUTING.md#modules) that everything depject is to be considered deprecated, so instead of working out how the f**k depject works, I thought I could just rip that out and `require` it from a new module.
Of course, I'm not sure if I put the new file in the right place. Open to suggestions here.
It works for me though, so...

black-puppydog:
Okay, one tipsy dev kinda found this okay. Has to be enough for now. Feel free to revert.

image

christianbundy:
> instead of working out how the f**k depject works, I thought I could just rip that out and require it from a new module

+1

> Okay, one tipsy dev kinda found this okay

+100

:rocket:

PR #1279, opened 2020-04-24
Handle travis cache more delicately.

black-puppydog:
After debugging in #1277, it seems that our travis cache got corrupted.
This simply puts a cache verification step in front of the build process.
If the caching becomes a problem now, we shoul at least get some clear
info on what's happening.

Also, this disables explicit caching of ~/.npm, because that is handled
implicitly by travis these days.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,6 @@ matrix:
 
 cache:
   directories:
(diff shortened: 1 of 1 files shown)


christianbundy:
I LOVE IT!

Issue #1285, opened 2020-05-05
Shortcut for publishing, when already in preview modal.


black-puppydog:
Brought up by `@G9kPTkgYZ0R+UxXDoppw/AcyHCZftD8NJbZda76CRYA=.ed25519` in `%n6isRKS9mZ9XltShMNRlmFNaU+AZj37Z8qLpMjJlcRw=.sha256`

PR #1286, opened 2020-05-05
Add publishing keyboard shortcuts

black-puppydog:
Allow fast publish and cancel actions when previewing posts.

Addresses #1285 by adding a key handler to the preview sheet.
It also put the focus on the cancel button. The result is that after
entering preview with `<Ctrl/Cmd> + <Return>` ...

* `<Return>` OR `<Escape>` will cancel the publish and go back to editing
* `<Ctrl/Cmd> + <Return>` once more will publish the message.


lib/depject/sheet/display.js
```diff
@@ -5,9 +5,13 @@ exports.gives = nest('sheet.display')
...
   return nest('sheet.display', function (handler) {
-    const { content, footer, classList, onMount } = handler(done)
+    const { content, footer, classList, onMount, attributes } = handler(done)
 
-    const container = h('div', { className: 'Sheet', classList }, [
+    let fullAttributes = { className: 'Sheet', classList }
+    if (attributes !== undefined) {
+      fullAttributes = { ...attributes, ...fullAttributes }
+    }
+    const container = h('div', fullAttributes, [
```
black-puppydog, line 14: This feels like it could be cleaner. I'm passing the `classList` and the `attributes` separately just to have them be recombined...

black-puppydog: Also, could I maybe attache the event handlers (passed as `attributes`) to something else that makes more sense?

christianbundy:
Concern: I know that Microsoft Word

black-puppydog:
I wouldn't target feature parity with MS Word, or even shared UX. :P

PR #1294, opened 2020-06-03
Dracula theme

black-puppydog:
This is by no means complete, but it seems usable to me. And hey, it's a config option, so no harm in adding it I guess?
The thing that bugs me the most at the moment is the orange border around the filter input (last screenshot)

screenshot of new patchwork dracula theme; public timeline

screenshot of profile page in dracula theme

profile list sheet with dracula theme


christianbundy:
BEEP BOOP MERGED

Thanks for making this, it's fun!

PR #1304, opened 2020-07-16
Ignore whitespace-only ContentWarnings.

### LINKED ISSUE #1303: Content warning should be trimmed
Noticed after publishing `%6LDp9KsZakipOaC5C3mw4wtMmM/qhm49ljJvaXtD35Y=.sha256` which has CW `" "`

black-puppydog:
This will break the workflow of people that use `"    "` or such as a CW
that means "beware, but I won't tell you why". :)

closes #1303 

I also thought to treat whitespace-only message body (in the compose input) as empty, and disable the publish button in that case. But that felt like a different issue.

### DIFF
--- a/lib/depject/message/html/compose.js
+++ b/lib/depject/message/html/compose.js
@@ -278,8 +278,9 @@ exports.create = function (api) {
         })
       })
 
-      if (contentWarningInput.value) {
-        content.contentWarning = contentWarningInput.value
+      const cw = contentWarningInput.value.trim()
+      if (cw.length > 0) {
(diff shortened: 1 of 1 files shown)


cryptix:
SGTM

I don't think there is much point in an empty CW

cryptix:
Code wise I'd propose to pull the value out and then check if it's longer then zero. Something like:

```js
let cw = contentWarningInput.value.trim()
 if (cw.length > 0) {
        content.contentWarning = cw
}
```

PR #1335, opened 2020-12-16
bump mutant to stable version 3.29.0

jedahan:


black-puppydog:
Thanks for pointing it out @jedahan! This works fine for me locally.

Issue #1037, opened 2019-05-13
Switching from "most active" to "recent" in tags freezes app for several minutes on OSX


HendrikPetertje:
Hey there, Lemme first say that I absolutely love the Patchwork app and appreciate the enormous effort all of you have put in it so far!
I'll describe my problem in some kind of readable format:

re-produce:
- Click on "more>" menu in the top and hit "tags",
- In the tags window click on the tab "From Everyone"
- A spinner shows up in the top of the screen with "Scuttling", the app remains usable and clickable during this "scuttling".
- In the side-bar I switch from "Most Active" to "Recent"
- The interface freezes for 2 minutes or so, I can't navigate and the cursor remains a pointer.
- After 2 minutes the Recent tab becomes highlighted and I see the "Scuttling" animation at the top of the screen.
- Posts start to load.

What I expected:
Clicking Recent not freezing the interface

What happend:
The interface frooze.

Verison information:
- OSX Mojave, version 10.14.4 on a Macbook pro (15-inch 2017)
- Patchwork v3.12.0-beta.2 (happens on Patchwork v3.12.0-beta.0 too)
- gossip.json size: 393kb
- total SSB size: 7.1 gb

HendrikPetertje:
tabbing back to "most active" yields the same results.

christianbundy:
Thanks for the awesome bug report! My intuition is that there's probably a bottleneck happening in scuttle-tag or the code around it. The code in #887 *might* fix this, but I haven't had time to test that branch.

black-puppydog:
Re-opening this because in the course of debugging #1337 I stumbled over it.
For me this happens both on `master`, on the two most recent AppImages, and on that PR.
And even though I have left it in that state for long times (hours) it has never finished.
However, it only happens with my main identity. With smaller test identities it also doesn't happen on any of the aforementioned versions.

Since I didn't know about this issue it took me some time to figure this out; I thought something had silently crashed.
In the end, I decided to limit the number of tags involved, simply by adding `.slice(0, 10)` to this line:

https://github.com/ssbc/scuttle-tag/blob/cc007759a338d67035f7a68e42a054b6a36d8143/obs.js#L214

That "fixes" the issue. So it seems like nothing is fundamentally broken, but there is definitely a performance issue. I suspect that some operation has a polynomial factor we didn't realize, otherwise it would work (albeit very slowly) for my main identity.

black-puppydog:
[...]

PR #1337, opened 2021-01-09
Version bumps

black-puppydog:
First batch of mostly no-op version bumps.
I.e. the bumped versions mostly add bugfixes, docs, tests etc to the dependencies themselves, but don't change behaviour.

The only bump that I had to actually change anything for was the bump of scuttle-tags, which changed its API a bit. 
The tags were unreliable for me before already, but started working after I deleted `~/.ssb/flume/tags.json`. After that, the version bump didn't break them again.
So there might be a separate issue where the index for tags is broken or at least gets stuck somehow, but it seems to be orthogonal to the version bump.

As I described on ssb, I'm going about this very slowly, partly due to hardware issues, and partly because we don't have tests (except `standard` which doesn't test anything).
I saw that @Powersource (?) had a WIP PR for that at some point, but I guess that's a herculean task, so didn't really go anywhere.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,12 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://help.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  - package-ecosystem: "npm" # See documentation for possible values
+    directory: "/" # Location of package manifests
+    schedule:
+      interval: "daily"
+    open-pull-requests-limit: 16
--- a/lib/depject/page/html/render/tag.js
+++ b/lib/depject/page/html/render/tag.js
@@ -44,17 +44,17 @@ exports.create = function (api) {
     const taggedMessages = MutantArray([])
     let messageStream
     if (tagId === 'all' && author === 'all') {
-      messageStream = scuttleTag.pull.messagesTagged()
+      messageStream = scuttleTag.pull.allTags()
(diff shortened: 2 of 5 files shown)


Powersource:
If you had to delete tags.json to get stuff to work then we probably shouldn't merge that update.

black-puppydog:
No, I had to delete tags.json to get tags to work *before* the version bump.
But of course, if people encounter problems with this PR, then yeah, no can do. :P

I think to facilitate this whole process, I'd like to see if we can just build an AppImage for every PR and attach it here automatically. That would serve as a first test whether the thing even builds, and also to more quickly test the PRs.

Powersource:
Btw wow #1337

black-puppydog:
Ah, worth noting is also that this would enable dependabot here. Not sure if we want this. I could instead keep that on my fork for now until the dust settles a bit.

black-puppydog:
Okay I tested this again with the last appimage and in fact, both the last release and this PR have the same issue:
Steps to reproduce:

1. Open patchwork, click on "More > Tags". The list of tags populates (can take a while if not used recently I guess) and eventually they are shown by name instead of hash prefic.
2. Click on "Recent" in the left sidebar.

At this point, patchwork seems to freeze without any client- or server-side console output.
But as I said, this happens on both versions for me. If someone can reproduce this on the latest release, I'd file an issue and merge this PR anyhow.

black-puppydog:
Okay, it really seems the tags thing is unrelated. See #1037 for details

black-puppydog:
Okay, lacking testing from others here, but having run this for a few days now, I think this is good.
Plan from here is to keep grinding through the current & to-come bump PRs, and then make a beta release so even non-technical folks could test it.

PR #1338, opened 2021-01-10
Flumeview reduce periodic

black-puppydog:
Wouldn't merge this one just yet, since I just filed the upstream PR. Let's see how that goes, I'd prefer to just bump upstream.
I'll just leave this here more as a reminder until https://github.com/flumedb/flumeview-reduce/pull/16/ is addressed one way or another. 

If anyone wants to test this, here's how you can see that this works:

1. Close patchwork
2. Delete a view that uses `flumeview-reduce`. For example: `rm ~/.ssb/flume/patchwork-contacts.json`
3. Start patchwork, let it run for a little. 
4. Check with `ssb-server status | jq .sync.plugins` that the contacts view is updating.
5. Stop patchwork (with Ctrl-C or by closing the window) *before* it catches up with the other views.
6. Restart patchwork
7. Check that the view does not restart from scratch

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,12 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://help.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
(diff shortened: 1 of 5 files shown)


black-puppydog:
(needless to say, all the other commits should disappear if/when that other bump PR is merged)

black-puppydog:
Christian merged this and already bumped the flumeview-reduce version, so I'll join this PR with the version bumps PR. <3

PR #1357, opened 2021-01-12
Version bumps take 2

black-puppydog:
Another round of ssb unrelated bumps. 
I'll leave these here to stew for a bit in case anyone wants to weigh in (and because I'm dead tired by now so shouldn't merge anything into anything).
Otherwise will self-merge again and leave testing for the next beta.


black-puppydog:
Been patchworking all day long, this is working great! Merging! :stuck_out_tongue:

PR #1366, opened 2021-01-13
ssb dependency version bumps: the easy parts

black-puppydog:
These are the version bumps that are no-ops or trivial, or are themselves mostly docs, tests, and version bumps themselves.

This will recompute some indexes due to changes in their dependencies:

- `ssb-private`, `ssb-friends`, and `ssb-query` bump `flumeview-query`
- `ssb-backlinks` bumps `flumeview-links`

So a bit of reindexing time is expected, which means patchwork will start up but then look quite empty and use a bunch of CPU & I/O for a while.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4116,6 +4116,113 @@
         "pull-write": "^1.1.1"
       }
     },
+    "flumeview-links": {
+      "version": "2.1.0",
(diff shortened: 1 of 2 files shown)


black-puppydog:
Have been running this all day (and last night) again. Merging. We'll get there, just you wait! :)

PR #1369, opened 2021-01-15
[WIP] display blocking reasons

black-puppydog:
This is not beautiful yet, but it already works.
The idea is to render block reasons a full markdown-capable messages like posts. The reasoning being that I don't want to force people to be terse when blocking. (of course they still can be, to the point of not giving a reason)

**Issues:**

The html that this generates doesn't really match that of posts (which also have a different heading) and so it looks bad

Also this reuses the `Message` element and so cannot be styled differently from a normal message. Not sure if we need that though. Frankly I think the messages integrate well if they render exactly like (other) messages. Open for discussion on that though.

### DIFF
--- a/lib/depject/feed/html/meta-summary.js
+++ b/lib/depject/feed/html/meta-summary.js
@@ -121,7 +121,7 @@ function getActions (msgs) {
     if (content.type === 'contact') {
       if (ref.isFeed(content.contact)) {
         const to = content.contact
-        const key = `${from}:${to}`
+        const key = `subscribe:${from}:${to}`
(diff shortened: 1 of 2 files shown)


black-puppydog:
kas mentioned that prior to using `reason` as the... reason... to block, they used `comment`.
So if we don't find `reason` but do find `comment` I'd render that.

black-puppydog:
Wow I'm so sorry... Imagine you manage to get & hold `@x` on a platform like github, and then it turns out to be an absolute nightmare due to idiots like me. :see_no_evil:

PR #1372, opened 2021-01-16
electron: 7.2.4 -> 8.5.5

fabianhjr:
Release Notes: https://www.electronjs.org/releases/stable?version=7 & https://www.electronjs.org/releases/stable?version=8

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2688,9 +2688,9 @@
       }
     },
     "electron": {
-      "version": "7.2.4",
-      "resolved": "https://registry.npmjs.org/electron/-/electron-7.2.4.tgz",
-      "integrity": "sha512-Z+R692uTzXgP8AHrabE+kkrMlQJ6pnAYoINenwj9QSqaD2YbO8IuXU9DMCcUY0+VpA91ee09wFZJNUKYPMnCKg==",
+      "version": "8.5.5",
+      "resolved": "https://registry.npmjs.org/electron/-/electron-8.5.5.tgz",
+      "integrity": "sha512-e355H+tRDial0m+X2v+l+0SnaATAPw4sNjv9qmdk/6MJz/glteVJwVJEnxTjPfEELIJSChrBWDBVpjdDvoBF4Q==",
       "dev": true,
       "requires": {
         "@electron/get": "^1.0.1",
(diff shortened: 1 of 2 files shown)


black-puppydog:
Hi @fabianhjr, thanks for the nudge :)
Yes, bumping electron is on my todo list. Really I'd want to bump it to the last active version over time, but starting small is of course a good start, especially since I'm not an electron expert. :P
But we need to bump to at least v9 to even still be in the range of releases that receive security fixes. 
From my understanding only the last three versions get those, and currently v11 is the latest stable version, with a new one hitting every three months. 

Fun! :D

fabianhjr:
There were no warnings on 7.2.4 -> 7.3.3, and 1 warning on 7.3.3 -> 8.5.5

`(electron) The default value of app.allowRendererProcessReuse is deprecated, it is currently "false".  It will change to be "true" in Electron 9.  For more information please check https://github.com/electron/electron/issues/18397`

Patchwork is working fine on 8.5.5.

black-puppydog:
Wowowow, you're not waiting around, eh? :)
Great to know!
Since my item of "push a beta" seems to be stalled for today, I might just as well verify this works for me too, and merge if so.

fabianhjr:
Electron 9 doesn't currently work:

```
(node:28397) UnhandledPromiseRejectionWarning: Error: An object could not be cloned.
    at EventEmitter.t.ipcRendererInternal.send.t.ipcRendererInternal.send (electron/js2c/renderer_init.js:127:332)
    at electron/js2c/renderer_init.js:123:343
(node:28397) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:28397) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
ssb-friends: stream legacy api used
Error: Loading non-context-aware native module in renderer: '/home/fabian/Development/Open/patchwork/node_modules/@felixrieseberg/spellchecker/build/Release/spellchecker.node', but app.allowRendererProcessReuse is true. See https://github.com/electron/electron/issues/18397.
```

black-puppydog:
Okay yes, version 8 works for me. Epic!
As for that warning: we don't use `allowRendererProcessReuse` in patchwork, and grepping for it (i.e. including in dependencies) comes up empty for me.
The discussion in the [issue](https://github.com/electron/electron/issues/18397) mentioned in the warning is actually helpful in understanding the source. 
The "workaround" for now could be to add `electron.app.allowRendererProcessReuse = false` to `index.js` just after the `require()` for electron.

black-puppydog:
But long-term, we probably want to make sure our native dependencies are context-aware (which *of course* I don't understand the implications of)

hjacobs:
I just tried out this version bump (Electron 8.5.5) and it seems to work fine for me. Should we merge @black-puppydog or is there any reason not to update?

black-puppydog:
@hjacobs I was mostly just held up with getting GH actions running and such. It's pretty relaxing even to just have some smoke testing going on.

PR #1381, opened 2021-01-24
Revert ssb-keys-neon for now, bump ssb-keys & scuttle-blog

black-puppydog:
For me this removes a bunch of dependency duplication too.


black-puppydog:
Argh, here all builds worked, and after merging one of them fails. wtf windows, wtf? :neutral_face:

PR #1399, opened 2021-02-02
Remove depject, epic 1: shortFeedId

### LINKED ISSUE #1371: Deprecate depject (PoC)
The docs say depject is considered deprecated. 
I just had to read into it yesterday for the first time, and Dominic's rationale behind having it work the way it does, seems to have been that he didn't want to exert undue control over the course of ssb. It could be summarized as "opinionatedly unopinionated".

Patchwork is no longer the only (or even the primary) ssb client. I think it is fine now to have opinions on code in the repo, especially given that what we're short on is not consensus but brain cycles.

The end goal here is to get rid of depject altogether and replace it with simple imports where possible. But as a total depject (and still kinda JS) noob, I won't go for that directly.

As a PoC, I want to choose any single depject module, and refactor it into a regular JS file that then gets `require`d from elsewhere.
Ideally this would be a "leaf" (only `gives`, no `needs`) so that this is easy.
Next would be a single, but hopefully simple/small `needs` & `gives` module to see how that goes. 

And then draw the rest of the f**king dog. :stuck_out_tongue:

black-puppydog:
This removes the first file from claws of depject. Well, *my* first file anyhow.
This closes #1371, but *really* it is just the start of an epic battle. :laughing: 

@christianbundy is there any chance you'd remember why there were two near-identical functions with the same name? 
The one I removed added "..." to the end of the shortened feedIds, but it didn't get used anywhere it seems. Was that by accident?
In any case, if some code would want the "..." then I'd say that can be done outside this function. 
Frankly that function might be inlined entirely, but it was a good punching ball for me.

### DIFF
--- a/lib/depject/about/obs.js
+++ b/lib/depject/about/obs.js
@@ -1,3 +1,4 @@
+const shortFeedId = require('./sync.js')
 const { computed } = require('mutant')
 const nest = require('depnest')
 const ref = require('ssb-ref')
@@ -9,7 +10,6 @@ const MutantPullDict = require('../../mutant-pull-dict')
 exports.needs = nest({
   'sbot.pull.stream': 'first',
   'blob.sync.url': 'first',
-  'about.sync.shortFeedId': 'first',
   'keys.sync.id': 'first'
 })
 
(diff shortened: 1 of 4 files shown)


christianbundy:
Nice! 🎉 

I wish I remembered. The only thing that sticks out to me is that the "<id>..." pattern was commonly used with SSB-Markdown to render long IDs as short IDs. But otherwise, I got nothing. 🤷

black-puppydog:
Well, it hasn't mattered in a while I guess. :D
And *nobody* complained about the absence of ellipsis. So... :shrug:

PR #1418, opened 2021-02-27
version bumps and cleanups

black-puppydog:
Already tested locally. let's hope this builds on all the node & os versions :P


black-puppydog:
Excellent!

PR #1419, opened 2021-02-28
Electron bump

black-puppydog:
This is very much WIP, but maybe folks want to test it...?

This is still lacking the context menu and for some reason the completion boxes (in search bar and post textareas) are slow as hell.
Apart from that it seems to work... 🙂


mwmiller (commented):
LGTM with minimal oversight

black-puppydog:
Couldn't help myself and debugged a bit more.
The idea right now is to use electron-context-menu to replace electron-spellchecker. It should actually make this quite easy to do.
However, what I see is that when I right-click somewhere, the context menu only flickers in, then out immediately, so fast that all I can see is a white blip, and often not even that.
The building handler for the menu is called though. The menu just emits a `menu-will-close` event right afterwards.

lib/window.js
```diff
@@ -6,35 +6,28 @@ module.exports = function Window (config, path, opts) {
   const window = new electron.BrowserWindow(extend({
     show: false,
     webPreferences: {
-      nodeIntegration: true // XXX: Maybe not always necessary (?)
+      nodeIntegration: true, // XXX: Maybe not always necessary (?)
+      enableRemoteModule: true,
```
black-puppydog, line 10: The goal is to be able to remove this line again.

lib/window.js
```diff
@@ -6,35 +6,28 @@ module.exports = function Window (config, path, opts) {
...
-
-      document.documentElement.replaceChild(h('body', [
-        rootView(config, data)
-      ]), document.body)
-    `)
+    window.webContents.send('window-setup', {
+      rootPath: path,
+      config: config,
+      data: opts.data || '',
+      title: opts.title || 'Patchwork',
```
black-puppydog, line 29: If we can pass this somehow directly to the BrowserWindow upon creation, we can save a round-trip of ipc...

black-puppydog: Also (and this is not a new issue, but nevertheless) I am thinking to strip `config.keys.private` out of this payload, to prevent the renderer process from accessing private key material. Not only should that improve the security in case of e.g. remote code execution CVEs, but also it should remove a potential foot-gun by only having one point in the code being even *able* to sign new messages, thus potentially preventing bug-induced forks...?

lib/main-window.js
```diff
@@ -266,78 +257,99 @@ module.exports = function (config) {
...
-        click () {
-          navigate('/channels')
-        }
+        target: '/channels',
       }
     }
   }
 
-  function dropTab (title, items) {
+  function buildDropdownMenuItems() {
```
black-puppydog, line 289: In hindsight, this could be done in the main process.
But even then, I'd probably build it like this for legibility and keep the `buildMenu()` around to add the `navigate-to` event handlers afterwards. So... dunno, feels okay to have the description of the page structure in the renderer process, that's where the rest of it lives too, right?

